PR #5, opened 2026-03-31
Add mandatory patch version verification as Step 0

JuanVqz:
## Summary

- Adds a mandatory **Step 0: Verify Latest Patch Version** to the upgrade workflow
- Before hopping to the next minor/major Rails version, the skill now checks that the app is on the latest patch release of its current series (e.g., 3.2.19 → 3.2.22.5 before jumping to 4.0.x)
- Splits patch version lookup into **static table for EOL series** (≤ 6.1) and **dynamic RubyGems API lookups for active series** (≥ 7.0)
- Extracts version history into a standalone **CHANGELOG.md** and consolidates skill metadata into the Skill Identity section

## Why

The official Rails upgrade guide and community best practices (FastRuby.io, Shopify, GitHub) all recommend upgrading to the latest patch of the current series before attempting a minor/major version hop. Reasons:

- **Security fixes** in later patches that you'd carry as vulnerabilities otherwise
- **Bug fixes** that could cause confusing failures during upgrade, making it hard to distinguish pre-existing issues from upgrade-related ones
- **Additional deprecation warnings** in later patches that prepare for the next version
- **Smaller delta** — the jump to the next minor/major is better tested from the latest patch

The static/dynamic split prevents the patch versions table from going stale — EOL series are frozen forever, while active series are resolved at runtime via the RubyGems JSON API or `gem search`.

## Changes

- `SKILL.md`: Added Step 0 to core workflow, all request patterns, Key Principles, and Success Criteria. Consolidated metadata into Skill Identity section. Removed inline changelog.
- `reference/multi-hop-strategy.md`: Added "Step 0: Upgrade to Latest Patch Release" section with static EOL table and dynamic lookup commands for active series.
- `CHANGELOG.md`: New file — single source of truth for version history.

## Test plan

- [ ] Verify skill triggers Step 0 when app is not on latest patch
- [ ] Verify skill skips Step 0 when app is already on latest patch
- [ ] Verify EOL table values against rubygems.org
- [ ] Verify dynamic RubyGems API lookup returns correct latest patch for active series

### DIFF
--- a/rails-upgrade/SKILL.md
+++ b/rails-upgrade/SKILL.md
@@ -59,7 +59,9 @@ When proposing code fixes that must work with both the current and target Rails 
 ### Step 0: Verify Latest Patch Version (MANDATORY PRE-STEP)
 - **CRITICAL:** Before any upgrade work begins, verify the app is on the latest patch release of its current Rails series
 - Read `Gemfile.lock` to find the exact current Rails version (e.g., `3.2.19`)
-- Compare against the latest patch for that series (see `reference/multi-hop-strategy.md` for the reference table)
+- Compare against the latest patch for that series:
+  - **EOL series (≤ 6.1):** Use the static table in `reference/multi-hop-strategy.md`
+  - **Active series (≥ 7.0):** Query the RubyGems API at runtime (see `reference/multi-hop-strategy.md` for commands)
 - If the app is NOT on the latest patch:
   - Inform user: "Your app is on Rails X.Y.Z but the latest patch is X.Y.W — you should upgrade to the latest patch first"
   - Guide user through updating the Gemfile and running `bundle update rails`
@@ -235,8 +237,9 @@ When user requests an upgrade, follow this workflow:
 ⚠️  THIS STEP IS REQUIRED BEFORE ANY OTHER WORK
 
 1. Read Gemfile.lock to find exact current Rails version (e.g., 3.2.19)
-2. Compare against latest patch for that series
-   (see reference/multi-hop-strategy.md — "Latest Patch Versions Reference" table)
+2. Compare against latest patch for that series:
+   - EOL series (≤ 6.1): use static table in reference/multi-hop-strategy.md
+   - Active series (≥ 7.0): query RubyGems API (see reference/multi-hop-strategy.md for commands)
 3. If current version < latest patch:
    - INFORM user: "Your app is on Rails X.Y.Z but the latest patch is X.Y.W"
    - Guide through Gemfile update and bundle update rails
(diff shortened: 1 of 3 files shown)


rails-upgrade/reference/multi-hop-strategy.md
```diff
@@ -4,9 +4,67 @@
...
+| 5.1.x | 5.1.7 | End of life |
+| 5.2.x | 5.2.8.1 | End of life |
+| 6.0.x | 6.0.6.1 | End of life |
+| 6.1.x | 6.1.7.10 | End of life |
+| 7.0.x | 7.0.8.7 | Security only |
+| 7.1.x | 7.1.5.1 | Security only |
+| 7.2.x | 7.2.2.1 | Maintained |
+| 8.0.x | 8.0.2 | Maintained |
+
+> **Note:** This table may become outdated. Always verify against [rubygems.org](https://rubygems.org/gems/rails/versions) before starting an upgrade.
```
JuanVqz: If the table can become outdated, is it better if we remove it?

JuanVqz:
@etagwerker reviews are welcome.

I like this step because when doing upgrades myself, it pays off in three ways:
1. You gain confidence early — a patch upgrade is low-risk and fast, so you prove the upgrade workflow works before tackling the harder hop.
2. You start the next version jump from the best possible baseline — latest security patches, latest deprecation warnings, latest everything — which makes it easier and more informed.
3. From experience: I've seen breaking changes even within patch releases (I think it was Rails 4). Being on the latest patch meant debugging against a smaller delta rather than the full jump.

rails-upgrade/reference/multi-hop-strategy.md
etagwerker: @JuanVqz Yes, that's a good idea!

rails-upgrade/SKILL.md
```diff
@@ -483,17 +529,24 @@ A successful upgrade assistance session:
...
 **Skill Type:** Modular with external workflows and examples
 **Methodology:** Based on FastRuby.io upgrade best practices and "The Complete Guide to Upgrade Rails" ebook
 **Attribution:** Content based on "The Complete Guide to Upgrade Rails" by FastRuby.io (OmbuLabs)
 
+**v3.1 Changes:**
+- Added mandatory Step 0: Verify Latest Patch Version — ensures app is on latest patch of current series before any minor/major hop
+- Added latest patch versions reference table in `reference/multi-hop-strategy.md`
+- Updated workflow from 4-step to 5-step process
+- All request patterns now include Step 0 patch verification
+- Updated Key Principles and Success Criteria to include patch verification
```
etagwerker: @JuanVqz This should be moved to a CHANGELOG.md

etagwerker (approved):
@JuanVqz Looks good! A couple of small nitpicks: 

- Move version number to a separate file
- Add CHANGELOG.md to make it easier to keep track of what changed

rails-upgrade/reference/multi-hop-strategy.md
JuanVqz: Ended up splitting in two parts:
* static table for EOL versions since won't be updated (<= 6.1)
* Runtime lookup for active versions (=> 7.0)

rails-upgrade/SKILL.md
JuanVqz: Moved!

JuanVqz:
> @JuanVqz Looks good! A couple of small nitpicks:
> 
> * Move version number to a separate file
> * Add CHANGELOG.md to make it easier to keep track of what changed

Both are done. @etagwerker feel free to review again.

etagwerker (approved):

Issue #7, opened 2026-04-01
Incorrect order of steps regarding `load_defaults`


etagwerker:
The FastRuby process doesn't usually include this step this early: 

```
Step 2: Verify load_defaults Matches Current Rails Version

CRITICAL: Check if load_defaults in config/application.rb matches the current Rails gem version
DELEGATE to the rails-load-defaults skill for detection and incremental update
That skill handles tiered, per-config updates with test runs between each change
If load_defaults matches current Rails version → Proceed to Step 3
DEPENDENCY: Requires the [rails-load-defaults skill](https://github.com/ombulabs/claude-code_rails-load-defaults-skill)
```

It usually comes after the target version of Rails or Ruby are running in production without any problems. 

So we should adapt the skill to match our process as much as possible.

JuanVqz:
This one was addressed.  @arielj could you please double-check this? so we can close the issue

JuanVqz:
Has been solved

PR #6, opened 2026-03-31
Fix typos, version mismatches, and numbering errors

JuanVqz:
## Summary
- Fix step numbering gap in SKILL.md (steps jumped from 2 to 4)
- Fix Ruby version requirement for Rails 8.0 (3.1+ → 3.2+) across 3 files
- Fix duplicate/skipped numbering in upgrade-7.0-to-7.1.md and upgrade-8.0-to-8.1.md
- Fix extra hop label in multi-hop-upgrade.md diagram
- Update outdated "detection script" references to "direct detection" in upgrade-report-workflow.md

## Test plan
- [x] Review each file change for correctness
- [x] Verify step numbering is sequential
- [x] Confirm Ruby version requirements match official Rails docs

PR #8, opened 2026-04-01
Improve Rails 3.2 to 4.0 upgrade guide

etagwerker:
## Summary
- Added 4 new breaking changes from FastRuby.io field notes: `rescue_action` removal, partial magic variables removed, `cache_key` timestamp format change, and fixture date casting
- Fixed `rescue_from Exception` anti-pattern — examples now use specific exception classes
- Replaced `Gem::Version` idiom with `NextRails.next?` for dual-boot compatibility
- Reorganized priorities: moved `rescue_action` and partial magic variables from HIGH to MEDIUM

## Test plan
- [ ] Review new breaking change entries (#6, #7, #8, #14) for accuracy
- [ ] Verify numbering is sequential across HIGH (1-5), MEDIUM (6-13), LOW (14-25)
- [ ] Confirm Phase 3 checklist and Phase 5 testing sections reference the new items
- [ ] Confirm Common Issues section has matching entries for new breaking changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
+++ b/rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
@@ -92,10 +92,21 @@ class User < ActiveRecord::Base
 end
 ```
 
-**Temporary workaround (not recommended):**
+4. **Remove `require 'strong_parameters'`** (if backported from Rails 3.2):
+
+The `strong_parameters` gem is built into Rails 4.0. Any `require 'strong_parameters'` calls will fail if the gem is not in the Gemfile.
+
 ```ruby
-# Gemfile
-gem 'protected_attributes'
+# BEFORE
+require 'strong_parameters' # in engine.rb or controller
+
+# AFTER — remove the require entirely
+# (strong_parameters is built into Rails 4)
+```
+
+If you need dual-boot compatibility during the transition:
+```ruby
+require 'strong_parameters' unless NextRails.next?
 ```
 
 ---
@@ -287,7 +298,20 @@ has_many :invitations, :finder_sql => 'SELECT id from items where id is NULL'
 ```
 
 **Fix:**
-`:finder_sql` is deprecated in Rails 4 with no direct replacement. Rewrite using scopes, lambdas, or custom query methods.
+`:finder_sql` is deprecated in Rails 4 with no direct replacement. Rewrite using standard associations with scopes or custom query methods:
+
+```ruby
+# BEFORE
+has_many :invitations, :finder_sql => 'SELECT * FROM invitations WHERE invited_by_id = #{id}'
+
+# AFTER — rewrite as a standard association with a lambda
+has_many :invitations, -> { where('invited_by_id = ?', id) }
+
+# OR — if the SQL is too complex for a lambda, use a method
+def invitations
+  Invitation.find_by_sql(["SELECT * FROM invitations WHERE invited_by_id = ?", id])
+end
+```
 
 ---
 
@@ -316,11 +340,6 @@ User.find_by(name: name, email: email)
 User.find_or_create_by(email: email)
 ```
 
-**Temporary workaround:**
-```ruby
-# Gemfile
-gem 'activerecord-deprecated_finders'
-```
 
 ---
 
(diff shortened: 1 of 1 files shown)


rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -258,6 +537,263 @@ Rails 4.0 dropped support for `vendor/plugins`.
...
+# BEFORE
+accepted_at: <%= 3.days.ago %>
+
+# AFTER
+accepted_at: "<%= 3.days.ago.to_s(:db) %>"
+```
+
+---
+
+#### 15. `config.eager_load` Required in All Environments
```
JuanVqz, line 588: I'll check if we have more info about this one

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -258,6 +537,263 @@ Rails 4.0 dropped support for `vendor/plugins`.
...
+# BEFORE
+request.env.merge!(headers)
+
+# AFTER
+request.headers.merge!(headers)
+```
+
+---
+
+#### 23. `strong_parameters` Gem Built-in
```
JuanVqz: - Is this part of number 2 (strong params)?
- Can we merge them together?

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -100,32 +100,195 @@ gem 'protected_attributes'
...
+
+##### 3i. `:finder_sql` deprecated
+
+**Detection Pattern:**
+```ruby
+has_many :invitations, :finder_sql => 'SELECT id from items where id is NULL'
+```
+
+**Fix:**
+`:finder_sql` is deprecated in Rails 4 with no direct replacement. Rewrite using scopes, lambdas, or custom query methods.
```
JuanVqz: Do we have real examples? Let's add some to make it easier for the robots.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -188,7 +351,123 @@ get '/home' => 'home#index'
...
+<%= render partial: "something", locals: { something: nil } %>
+```
+
+No fix is needed when rendering with `collection:` or `object:` — the variable is still defined in those cases.
+
+**Detection Script:**
+```ruby
+require 'find'
+
+SEARCH_DIR = 'app/views'
```
JuanVqz, line 434: Will this cover packs, engines views?

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -188,7 +351,123 @@ get '/home' => 'home#index'
...
+```ruby
+# Code that stores or compares cache_key strings
+cache_key
+cache_timestamp_format
+```
+
+**Fix:**
+If your code stores cache keys externally (e.g., in Redis, a database, or a background job), those stored keys will no longer match after the upgrade. Either:
+1. Invalidate/regenerate stored cache keys after upgrading
+2. Or set `self.cache_timestamp_format = :number` on affected models to preserve the old format
```
JuanVqz, line 485: Given this has two possible fixes, Do you know what the skill will do? Ask us?

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
JuanVqz, line 512: Is this snippet correct? Is the same gem for the previous step `9. Observers Extracted`

JuanVqz: - Do we want to encourage the workarounds?
- Or it should ask?

To me we should do the actual changes, because is a trivial fix for the bots.

etagwerker: Good point

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
etagwerker: I don't think it will. If we want to add support for packs, I think it should be a different issue to address in the future.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
etagwerker: I believe the skill will ask the user to pick one or the other.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
etagwerker: Yes, rails-observers is correct. The gem bundles both Observers and Sweepers.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
etagwerker: Good idea

PR #10, opened 2026-04-06
Fix code errors, extract script, and deduplicate common issues

JuanVqz:
## Summary

Review fixes for PR #8 (Improve Rails 3.2 to 4.0 upgrade guide):

- **Fix `extending()` syntax** in section 3c — `(extending Foo)` is not valid Ruby, changed to `where(...).extending(Foo)`
- **Fix `finder_sql` lambda** in section 3i — bare `id` inside a lambda refers to the relation, not the owner; changed to `->(owner) { where(invited_by_id: owner.id) }`
- **Extract detection script** for partial magic variables to `detection-scripts/detect-partial-magic-vars.rb`, replacing 25 inline lines with a one-line reference
- **Remove extra blank line** after Dynamic Finders section
- **Deduplicate Common Issues** — replaced 12 entries (which repeated content from Breaking Changes) with a compact error→section lookup table
- **Replace `§` symbols** with plain "Section N" text for LLM-friendliness

Reduces the file from ~1025 to 884 lines.

## Test plan

- [x] Verify code examples in sections 3c and 3i are valid Ruby
- [x] Verify detection script runs: `ruby detection-scripts/detect-partial-magic-vars.rb`
- [x] Verify Common Issues table references match the correct section numbers


rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -185,9 +185,9 @@ has_one :active_visit, class_name: "Visit",
 belongs_to :clinic_patient_link, primary_key: :person_id, foreign_key: :person_id,
   conditions: clinic_id_conditions_proc, extend: MultiKeyAssociation::BelongsTo
 
-# AFTER
+# AFTER — this pattern is rare and complex; verify manually
 belongs_to :clinic_patient_link, ->(object) {
-  (extending MultiKeyAssociation::BelongsTo).where(clinic_id_conditions_proc.call(object))
```
JuanVqz, line 190: This feels a bit inconsistent. Claude suggested this syntax, but during code review it flagged it as invalid.

A few things are going on here:

* **`belongs_to` scopes don’t receive the owner object**
  In Rails 4, only `has_many` and `has_one` associations pass the owner into the lambda. A `belongs_to` scope is just a plain `-> { }`, so it has no access to the instance.

* **`(extending Foo)` is not valid syntax**
  There’s no receiver in that expression. `extending` needs to be called on a relation (e.g. `where(...).extending(Foo)`), not as a standalone expression.

* **Even though chaining `extending` is valid, it still won’t work here**
  Something like `-> { where(clinic_id: ...).extending(Foo) }` is syntactically correct. However, within a `belongs_to` scope you don’t have access to the owner instance, so you can’t evaluate something like `clinic_id_conditions_proc.call(object)` anyway.

So the issue here isn’t just syntax—it’s also a limitation of how `belongs_to` scopes work in Rails 4.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -305,7 +305,9 @@ has_many :invitations, :finder_sql => 'SELECT id from items where id is NULL'
 has_many :invitations, :finder_sql => 'SELECT * FROM invitations WHERE invited_by_id = #{id}'
 
 # AFTER — rewrite as a standard association with a lambda
-has_many :invitations, -> { where('invited_by_id = ?', id) }
+# Note: the owner must be passed as a parameter, since `id` inside a
+# bare lambda refers to the relation scope, not the owning record.
+has_many :invitations, ->(owner) { where(invited_by_id: owner.id) }
```
JuanVqz, line 310: `id` inside a lambda scope refers to the relation, not the owning
record. The previous example `-> { where('invited_by_id = ?', id) }`
would produce incorrect SQL or an error. Changed to use the owner
parameter: `->(owner) { where(invited_by_id: owner.id) }`, consistent
with the pattern shown in section 3c.

rails-upgrade/version-guides/upgrade-3.2-to-4.0.md
```diff
@@ -856,138 +856,24 @@ For each model with `attr_accessible`:
 
 ---
 
-## Common Issues
```
JuanVqz, line 886: Every Common Issues entry except Mass Assignment was a shorter copy of
content already present in the Breaking Changes sections above.

Replaced with a compact table that maps error messages to the relevant
section number, so the agent can quickly cross-reference when it
encounters an error during an upgrade.

JuanVqz (commented):
@etagwerker I opened a PR on top of your branch so we can continue working asynchronously.

Could you take a look at the review changes and confirm they look correct to you? If everything looks good, we can merge those into your PR after that, I’ll go ahead and merge it.

etagwerker (approved):

PR #23, opened 2026-04-09
Add detection patterns for Rails 4.2 → 5.0 upgrade

### LINKED ISSUE #13: Add detection patterns for Rails 4.2 → 5.0 upgrade
Detection patterns exist for Rails 6.0+ upgrades but are missing for older upgrade paths.

The `version-guides/upgrade-4.2-to-5.0.md` guide exists but there is no corresponding `detection-scripts/patterns/rails-50-patterns.yml` file.

## Acceptance criteria

- [ ] Create `rails-50-patterns.yml` covering all breaking changes from the version guide
- [ ] Organize patterns into `high_priority` and `medium_priority` sections
- [ ] Include `dependencies` section for any bridge/compatibility gems
- [ ] Verify YAML parses without errors
- [ ] Verify naming convention matches `rails-{VERSION}-patterns.yml` for auto-discovery by `direct-detection-workflow.md`

etagwerker:
## Summary
- Add `rails-50-patterns.yml` covering all breaking changes from `upgrade-4.2-to-5.0.md`
- 6 high-priority patterns: Ruby version, ApplicationRecord, belongs_to required, protected_attributes, attr_accessible, params as hash
- 5 medium-priority patterns: callback halt, assigns/assert_template in tests, UploadedFile, rake commands
- 2 dependency entries: rails-controller-testing, rails_upgrader

Closes #13

## Test plan
- [x] YAML parses without errors
- [x] Naming follows `rails-{VERSION}-patterns.yml` convention
- [ ] Verify patterns match the breaking changes in `version-guides/upgrade-4.2-to-5.0.md`
- [ ] Verify patterns are discoverable by `direct-detection-workflow.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
@@ -1,0 +1,126 @@
+version: "5.0"
+description: "Breaking change patterns for Rails 4.2 → 5.0 upgrade"
+
+breaking_changes:
+  high_priority:
+    - name: "Ruby version below 2.2.2"
+      pattern: "ruby.*['\"]2\\.[01]\\.|ruby.*['\"]1\\."
+      exclude: ""
+      search_paths:
+        - "Gemfile"
+        - ".ruby-version"
+      explanation: "Rails 5.0 requires Ruby 2.2.2 or newer"
+      fix: "Upgrade Ruby to 2.2.2+ (2.3+ recommended)"
+      variable_name: "RUBY_VERSION"
+
+    - name: "Models inheriting from ActiveRecord::Base"
+      pattern: "class\\s+\\w+\\s*<\\s*ActiveRecord::Base"
+      exclude: "ApplicationRecord"
+      search_paths:
+        - "app/models/"
+      explanation: "Rails 5.0 introduces ApplicationRecord as the base class for all models"
+      fix: "Create app/models/application_record.rb and change all models to inherit from ApplicationRecord"
+      variable_name: "ACTIVERECORD_BASE"
+
+    - name: "belongs_to without optional: true"
+      pattern: "belongs_to\\s+:\\w+"
+      exclude: "optional:"
+      search_paths:
+        - "app/models/"
+      explanation: "Rails 5.0 makes belongs_to required by default. Nullable associations will fail validation"
+      fix: "Add optional: true to belongs_to associations that can be nil"
+      variable_name: "BELONGS_TO_REQUIRED"
+
+    - name: "protected_attributes gem"
+      pattern: "gem.*['\"]protected_attributes['\"]"
+      exclude: ""
+      search_paths:
+        - "Gemfile"
+      explanation: "The protected_attributes gem is not compatible with Rails 5.0. You must migrate to Strong Parameters"
+      fix: "Remove protected_attributes gem, remove attr_accessible from models, add *_params methods to controllers"
+      variable_name: "PROTECTED_ATTRIBUTES"
+
+    - name: "attr_accessible in models"
+      pattern: "attr_accessible"
+      exclude: ""
+      search_paths:
+        - "app/models/"
+      explanation: "attr_accessible is removed in Rails 5.0. Use Strong Parameters in controllers instead"
+      fix: "Remove attr_accessible from models and add *_params methods to controllers"
+      variable_name: "ATTR_ACCESSIBLE"
+
+    - name: "Parameters used as hash"
+      pattern: "params\\[.*\\]\\.(slice|except|merge|symbolize_keys|to_hash)\\b"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+        - "app/"
+      explanation: "params is now ActionController::Parameters, not HashWithIndifferentAccess. Hash methods may not work"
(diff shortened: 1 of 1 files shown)


rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
```diff
@@ -0,0 +1,126 @@
...
+  medium_priority:
+    - name: "Callback returning false to halt"
+      pattern: "return\\s+false"
+      exclude: ""
+      search_paths:
+        - "app/models/"
+        - "app/controllers/"
+      explanation: "Returning false from a callback no longer halts the chain in Rails 5.0"
+      fix: "Replace 'return false' with 'throw :abort' in before_* callbacks"
+      variable_name: "CALLBACK_HALT"
```
JuanVqz, line 71: CALLBACK_HALT pattern is too broad. return false in app/models/ and app/controllers/ will match every method that returns false, not just callbacks. The  pattern should be more targeted, but realistically this is hard to narrow via regex alone. At minimum the explanation should clarify this is a noisy pattern that needs manual review, or the search_paths should be narrowed.

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
```diff
@@ -0,0 +1,126 @@
...
+      variable_name: "ACTIVERECORD_BASE"
+
+    - name: "belongs_to without optional: true"
+      pattern: "belongs_to\\s+:\\w+"
+      exclude: "optional:"
+      search_paths:
+        - "app/models/"
+      explanation: "Rails 5.0 makes belongs_to required by default. Nullable associations will fail validation"
+      fix: "Add optional: true to belongs_to associations that can be nil"
+      variable_name: "BELONGS_TO_REQUIRED"
```
JuanVqz, line 32: BELONGS_TO_REQUIRED will match every belongs_to. The pattern belongs_to\s+:\w+ with exclude optional: will flag every single belongs_to that doesn't already have optional: true, including ones that should be required. This is by design (it's a heads-up), but the explanation should say "review each belongs_to to determine if it should be optional" rather than implying they all need optional: true.

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
```diff
@@ -0,0 +1,126 @@
...
+
+    - name: "Parameters used as hash"
+      pattern: "params\\[.*\\]\\.(slice|except|merge|symbolize_keys|to_hash)\\b"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+        - "app/"
+      explanation: "params is now ActionController::Parameters, not HashWithIndifferentAccess. Hash methods may not work"
+      fix: "Call .permit and .to_h before using hash methods on params"
+      variable_name: "PARAMS_AS_HASH"
```
JuanVqz, line 60: PARAMS_AS_HASH pattern misses bare params.slice/params.except. The pattern requires params[.*] before the method call, so it won't catch params.slice(:name) or params.except(:id) directly, which is exactly what the version guide shows as detection examples (Section 5).

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
JuanVqz: Fixed
CALLBACK_HALT (commit af6d25b) - Added "This is a noisy pattern -- only matches inside before_* callbacks need fixing" to the explanation, and "Ignore  matches in non-callback methods" to the fix field.

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
JuanVqz: Fixed 
1. PARAMS_AS_HASH (commit 3078b74) - Changed the pattern from params\[.*\]\.(slice|...) to params(\[.*\])?\.(slice|...) so the bracket group is optional. Now catches both params.slice(:name) and params[:user].slice(:name).

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
JuanVqz: Fixed

BELONGS_TO_REQUIRED (commit 18a5b07) - Updated explanation to "Review each match to determine if the association can be nil" and fix to "Add optional: true  only to belongs_to associations that can legitimately be nil. Leave required ones as-is".

JuanVqz (approved):

rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml
```diff
@@ -0,0 +1,132 @@
...
+    check: true
+    gem_name: "rails-controller-testing"
+    message: "Required if tests use assigns() or assert_template"
+    url: "https://github.com/rails/rails-controller-testing"
+
+  test_unit:
+    check: false
+    gem_name: "test-unit"
+    message: "Required if you see 'cannot load such file -- test/unit/assertions' after upgrading Ruby"
+    url: "https://github.com/test-unit/test-unit"
```
JuanVqz, line 126: Added this missing one,  similar to `rails-controller-testing`

PR #28, opened 2026-04-17
Rename reference/ to references/ per Agent Skills spec

JuanVqz:
## Summary

Rename `rails-upgrade/reference/` to `rails-upgrade/references/` (plural) to match the [Agent Skills specification](https://agentskills.io/specification), which defines `references/` as the canonical optional directory for skill documentation.

The workspace `CLAUDE.md` states this repo follows the Agent Skills spec, so the directory name should align.

## Changes

- Directory renamed via `git mv` (history preserved for each file)
- Path references updated in:
  - `rails-upgrade/SKILL.md`
  - `rails-upgrade/workflows/upgrade-report-workflow.md`
- `rails-upgrade/CHANGELOG.md` left untouched — it records the path that existed at the time of that entry

## Why

Aligning the directory name with the canonical spec before the marketplace ships is cheaper than later, when external users may have bookmarked paths.


arielj (approved):
looks good

PR #30, opened 2026-04-17
Fix unknown env variables versions and more

arielj:
I'll add comments for each change in the diff. It's mostly updating EOL versions, fixing unknown variables, some formatting and some copy

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ Be aware of these limitations:
 | **Accurate time estimates** | The difficulty ratings and time estimates are rough guidelines based on typical applications. Your mileage will vary based on codebase size, test coverage, and custom code complexity. |
 | **Automated code changes** | The skill provides guidance and examples, but you implement the changes. It won't automatically refactor your code. |
 | **Gem compatibility resolution** | While we note common gem version requirements, resolving complex dependency conflicts requires manual intervention. |
+| **Rails LTS upgrades** | While many of the things this Skill can do will work to upgrade Rails LTS, the strategy for those apps will be different and the Rails source code is not the same as the main Rails repository |
 
 ## Contributing
 
--- a/rails-upgrade/SKILL.md
+++ b/rails-upgrade/SKILL.md
@@ -82,10 +82,10 @@ When proposing code fixes that must work with both the current and target Rails 
 ### Step 2: Set Up Dual-Boot with next_rails (EARLY SETUP)
 - **DELEGATE** to the `dual-boot` skill for setup and initialization
 - That skill handles:
-  - Checking if Gemfile.next already exists (to avoid duplicate next? method)
+  - Checking if Gemfile.next already exists (to avoid duplicate `next?` method)
   - Adding next_rails gem and running next_rails --init
   - Installing dependencies for both Rails versions
-  - Configuring the Gemfile with if next? conditionals
+  - Configuring the Gemfile with `if next?` conditionals
 - **DEPENDENCY:** Requires the [dual-boot skill](https://github.com/ombulabs/claude-code_dual-boot-skill)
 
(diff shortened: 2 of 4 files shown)


rails-upgrade/references/multi-hop-strategy.md
```diff
@@ -46,6 +46,8 @@ Before beginning any minor or major version hop, ensure you are on the **latest
 | 5.2.x | 5.2.8.1 |
 | 6.0.x | 6.0.6.1 |
 | 6.1.x | 6.1.7.10 |
+| 7.0.x | 7.0.10 |
+| 7.1.x | 7.1.6 |
```
arielj, line 50: these are also EOL now

rails-upgrade/references/multi-hop-strategy.md
```diff
@@ -248,7 +250,7 @@ For large applications, consider dual-boot testing:
 
 ```ruby
 # Gemfile
-if ENV['RAILS_NEXT']
+if next?
```
arielj, line 253: I think Claude hallucinated this env variable, we don't mention that anywhere else and it's not a next_rails thing either

rails-upgrade/SKILL.md
```diff
@@ -82,10 +82,10 @@ When proposing code fixes that must work with both the current and target Rails
 ### Step 2: Set Up Dual-Boot with next_rails (EARLY SETUP)
 - **DELEGATE** to the `dual-boot` skill for setup and initialization
 - That skill handles:
-  - Checking if Gemfile.next already exists (to avoid duplicate next? method)
+  - Checking if Gemfile.next already exists (to avoid duplicate `next?` method)
```
arielj, line 85: better formatting to match other places

rails-upgrade/SKILL.md
```diff
@@ -104,6 +104,7 @@ When proposing code fixes that must work with both the current and target Rails
 - Use `NextRails.next?` for code that must work with both versions (DELEGATE to dual-boot skill for patterns)
 - Update Gemfile to target Rails version
 - Run test suite against both versions during the transition
+- Do not fix deprecations printed by the next version, these will be addressed later before the next upgrade
```
arielj, line 107: this is an extra instruction because Claude would do extra changes that fix deprecations in the next version, adding more conditionals than needed (when these warnings can be addressed after the upgrade is done before the next version jump)

rails-upgrade/SKILL.md
```diff
@@ -382,7 +383,7 @@ Before starting ANY upgrade:
 - [ ] Plan merge strategy
 
 ### 5. Deprecation Warnings
-- [ ] Run app with `RAILS_DEPRECATION_WARNINGS=1`
+- [ ] Run app with Rails deprecations turned on (configured in config/environment files)
```
arielj, line 386: I think Claude hallucinated this env variable too

README.md
```diff
@@ -131,6 +131,7 @@ Be aware of these limitations:
 | **Accurate time estimates** | The difficulty ratings and time estimates are rough guidelines based on typical applications. Your mileage will vary based on codebase size, test coverage, and custom code complexity. |
 | **Automated code changes** | The skill provides guidance and examples, but you implement the changes. It won't automatically refactor your code. |
 | **Gem compatibility resolution** | While we note common gem version requirements, resolving complex dependency conflicts requires manual intervention. |
+| **Rails LTS upgrades** | While many of the things this Skill can do will work to upgrade Rails LTS, the strategy for those apps will be different and the Rails source code is not the same as the main Rails repository |
```
arielj, line 134: I think it's worth clarifying this here, as we have seen many Rails LTS projects and the strategy has to change, it's safer to set the expectation

JuanVqz (approved):
Looks good to me

PR #37, opened 2026-04-22
Add Rails 4.0 → 4.1 upgrade guide and detection patterns

### LINKED ISSUE #20: Add version guide and detection patterns for Rails 4.0 → 4.1 upgrade
The ebook ("The Complete Guide to Upgrade Rails", 5th Edition) covers Rails 4.0 → 4.1 in `from-4-0-to-4-1.md`, but the skill has no version guide or detection patterns for this upgrade path.

The SKILL.md upgrade paths table jumps from 3.2→4.0 to 4.2→5.0, skipping both 4.0→4.1 and 4.1→4.2. This contradicts the skill's core "never skip versions" principle.

This upgrade introduces Spring, secrets.yml, Action Mailer previews, and enums.

**Source material:** https://github.com/ombulabs/ebook/blob/main/from-4-0-to-4-1.md

## Acceptance criteria

- [x] Create `version-guides/upgrade-4.0-to-4.1.md` based on ebook content
- [x] Create `detection-scripts/patterns/rails-41-patterns.yml` covering all breaking changes
- [x] Verify YAML parses without errors
- [x] Add 4.0→4.1 row to SKILL.md supported upgrade paths table

JuanVqz:
## Summary

Closes #20.

- `version-guides/upgrade-4.0-to-4.1.md` — based on the OmbuLabs ebook chapter (`from-4-0-to-4-1.md`), the [official Rails 4.1 upgrade guide](https://guides.rubyonrails.org/v4.1/upgrading_ruby_on_rails.html#upgrading-from-rails-4-0-to-rails-4-1), and the 4.1 release notes. 20 breaking-change entries across HIGH/MEDIUM/LOW.
- `detection-scripts/patterns/rails-41-patterns.yml` — 22 matching detection patterns. YAML parses clean.
- `SKILL.md` — lists the new 4.0 → 4.1 guide (and the pre-existing 4.1 → 4.2 guide that was missing from the same list — see inline comment).

## Priority notes

- **MultiJSON → 🟡 MEDIUM** — not a hard boot break; raises `NameError` on first use and only affects apps that reference `MultiJSON` directly.
- **`return` in callbacks → 🔴 HIGH but inline-block only** — official docs confirm this raises `LocalJumpError` only for inline blocks; method-form callbacks with `return false` still work on 4.1. Scope is called out explicitly in the guide.
- **PG `json` / `hstore` string-keyed Hash → 🔴 HIGH** — silent runtime breakage for apps using these columns with symbol-key access.
- **`default_scope` chaining → 🟡 MEDIUM (tentative)** — candidate for HIGH per the CLAUDE.md "silently wrong behavior with production impact" rubric. Left at MEDIUM for now because it only hits apps that use `default_scope` in a contradicting way. Open to bumping.

## Coverage beyond the ebook chapter

The ebook chapter covers the AR/JSON/callback basics. The following items are additionally covered from the official 4.0 → 4.1 upgrade guide:

**HIGH**
- PostgreSQL `json` / `hstore` columns return string-keyed `Hash` (not `HashWithIndifferentAccess`)

**MEDIUM**
- Cookies serializer opt-in (Marshal → `:hybrid` / JSON)
- I18n `enforce_available_locales` defaults to `true`
- `as_json` millisecond precision for `Time` / `DateTime` / `TimeWithZone`

**LOW**
- `render :text` soft-deprecated in favor of `:plain` / `:html` / `:body`
- JSON encoder feature removals (circular detection, `encode_json` hook, BigDecimal-as-number) → `activesupport-json_encoder` gem
- JSON gem isolation from the Rails encoder (`JSON.generate` no longer honors `as_json`)
- Fixture ERB evaluated in a separate context (helpers via `ActiveRecord::FixtureSet.context_class`)
- `ActiveSupport::Callbacks.set_callback` around-block signature change
- Test helper `ActiveRecord::Migration.check_pending!` now redundant

## Test plan

- [x] `ruby -e "require 'yaml'; YAML.safe_load_file('rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml')"` succeeds
- [ ] Manual smoke: run the detection on a Rails 4.0 app and confirm the HIGH-priority patterns (dynamic finders, implicit join, inline-block callback return, PG json/hstore `store_accessor`) fire where expected
- [ ] Cross-check against RailsDiff 4.0.13 → 4.1.16 for anything missed

### DIFF
--- a/rails-upgrade/SKILL.md
+++ b/rails-upgrade/SKILL.md
@@ -204,6 +204,8 @@ If user requests a multi-hop upgrade (e.g., 5.2 → 8.1):
 
 **Legacy Rails:**
 - `version-guides/upgrade-3.2-to-4.0.md` - Rails 3.2 → 4.0 (Strong Parameters)
+- `version-guides/upgrade-4.0-to-4.1.md` - Rails 4.0 → 4.1 (Spring, secrets.yml, enums)
+- `version-guides/upgrade-4.1-to-4.2.md` - Rails 4.1 → 4.2 (ActiveJob, Web Console)
 - `version-guides/upgrade-4.2-to-5.0.md` - Rails 4.2 → 5.0 (ApplicationRecord)
 
 **Modern Rails:**
--- a/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
@@ -1,0 +1,145 @@
+version: "4.1"
+description: "Breaking change patterns for Rails 4.0 → 4.1 upgrade"
+
+breaking_changes:
+  high_priority:
+    - name: "MultiJSON usage"
+      pattern: "\\bMultiJSON\\b|require\\s+['\"]multi_json['\"]"
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+        - "config/"
+        - "Gemfile"
+      explanation: "Rails 4.1 no longer depends on MultiJSON. Apps that reference it directly will fail once Rails stops pulling it in transitively"
+      fix: "Either add gem 'multi_json' to the Gemfile, or migrate to core JSON: obj.to_json and JSON.parse(str)"
+      variable_name: "MULTI_JSON"
+
+    - name: "Dynamic finders (find_all_by / find_last_by / scoped_by)"
+      pattern: "\\.(find_all_by_|find_last_by_|scoped_by_)[a-zA-Z_0-9]+"
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 4.1 removed the activerecord-deprecated_finders dependency. find_all_by_*, find_last_by_*, and scoped_by_* no longer work"
+      fix: "Rewrite as where(col: value). For example: User.find_all_by_email(e) → User.where(email: e); User.find_last_by_email(e) → User.where(email: e).last. As a short-term bridge, add gem 'activerecord-deprecated_finders'"
+      variable_name: "DYNAMIC_FINDERS"
+
+    - name: "Dynamic find_or_initialize_by_* / find_or_create_by_*"
+      pattern: "\\.(find_or_initialize_by_|find_or_create_by_)[a-zA-Z_0-9]+"
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 4.1 removed the *_by_<col> forms of find_or_initialize_by and find_or_create_by"
+      fix: "Use the hash form: find_or_initialize_by(col: value) / find_or_create_by(col: value)"
+      variable_name: "FIND_OR_X_BY_DYNAMIC"
+
+    - name: "return from a callback block"
+      pattern: "^\\s*(before_|after_|around_)(save|create|update|destroy|validation|commit|rollback|initialize|find|touch)\\b.*\\{[^}]*\\breturn\\b"
+      exclude: ""
+      search_paths:
+        - "app/models/"
+        - "app/controllers/"
+        - "lib/"
+      explanation: "Rails 4.1 no longer supports returning from callbacks (including `return false`) as a way to halt the chain"
+      fix: "Replace `return false` with a bare `false` expression. In Rails 5+ this becomes `throw :abort`"
+      variable_name: "CALLBACK_RETURN"
+
+    - name: "Implicit join reference in includes+where string"
+      pattern: "\\.includes\\([^)]+\\)[^\\n]*\\.where\\(\\s*['\"][a-z_][a-z_0-9]*\\."
+      exclude: "\\.references\\(|\\.eager_load\\("
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 4.1 removed the string-parsing heuristic that auto-joined tables referenced inside a where(\"other_table.col = ?\") string after includes(...)"
+      fix: "Replace with explicit joins(:assoc) (no eager load), eager_load(:assoc) (eager loaded), or includes(:assoc).where(...).references(:assoc)"
+      variable_name: "IMPLICIT_JOIN_REFERENCES"
+
+    - name: "disable_implicit_join_references config (dead setting)"
+      pattern: "disable_implicit_join_references"
+      exclude: ""
+      search_paths:
+        - "config/"
+      explanation: "config.active_record.disable_implicit_join_references does nothing in Rails 4.1 and emits a deprecation warning"
+      fix: "Remove config.active_record.disable_implicit_join_references from your config files"
+      variable_name: "DISABLE_IMPLICIT_JOIN_REFERENCES"
+
+  medium_priority:
+    - name: "default_scope in models"
+      pattern: "^\\s*default_scope\\b"
+      exclude: ""
+      search_paths:
+        - "app/models/"
+      explanation: "Rails 4.1 chains default_scope with subsequent scopes (AND) instead of letting them override. Scopes that contradict the default can now return zero rows"
+      fix: "Audit each scope on models with default_scope. Use unscope(where: :col) or rewhere(col: value) where a scope is supposed to override the default"
+      variable_name: "DEFAULT_SCOPE"
+
+    - name: "ActiveRecord::Relation mutator calls"
+      pattern: "\\.where\\([^)]*\\)[^\\n]*\\.(compact!|map!|delete_if|reject!|select!|flatten!|sort!|uniq!)"
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 4.1 removed mutator methods from ActiveRecord::Relation. Calling compact!/map!/delete_if/etc. on a Relation raises"
+      fix: "Call .to_a first, then mutate the array: records = Project.where(...).to_a; records.compact!"
+      variable_name: "RELATION_MUTATORS"
+
+    - name: "Controller test with format: :js (should use xhr)"
+      pattern: "^\\s*(get|post|put|patch|delete)\\s+:[a-z_]+[^\\n]*format:\\s*:js"
+      exclude: "\\bxhr\\b"
+      search_paths:
+        - "test/controllers/"
+        - "test/functional/"
+        - "spec/controllers/"
+      explanation: "Rails 4.1 extends CSRF protection to GET requests with JS responses. Controller tests that hit JS endpoints with the plain verb helpers now fail CSRF or skip the XHR code path"
+      fix: "Use xhr :verb, :action, params instead of verb :action, format: :js"
+      variable_name: "JS_CONTROLLER_TEST"
+
+    - name: "flash.to_hash with symbol key filters"
+      pattern: "flash\\.to_hash\\.(except|slice|values_at)\\(\\s*:"
+      exclude: ""
+      search_paths:
+        - "app/"
+      explanation: "flash.to_hash keys are strings in Rails 4.1. Filtering with symbol keys silently no-ops"
+      fix: "Use string keys: flash.to_hash.except(\"notify\") instead of flash.to_hash.except(:notify)"
+      variable_name: "FLASH_SYMBOL_KEYS"
+
+  low_priority:
+    - name: "Spring not in Gemfile"
+      pattern: "source\\s+['\"]https?://rubygems"
+      exclude: ""
+      search_paths:
+        - "Gemfile"
+      explanation: "FILE MARKER (manual review required): Rails 4.1 ships with Spring in the default Gemfile. Apps generated before 4.1 will not have it"
+      fix: "Add gem 'spring' under group :development in the Gemfile, then run `bundle exec spring binstub --all` to generate Spring-aware binstubs. Optional but recommended"
+      variable_name: "SPRING"
+
+    - name: "secrets.yml missing"
+      pattern: "Rails\\.application\\.config\\.secret_(key_base|token)"
+      exclude: ""
+      search_paths:
+        - "config/"
+        - "app/"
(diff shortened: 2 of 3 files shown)


rails-upgrade/SKILL.md
```diff
@@ -203,6 +203,8 @@ If user requests a multi-hop upgrade (e.g., 5.2 → 8.1):
 
 **Legacy Rails:**
 - `version-guides/upgrade-3.2-to-4.0.md` - Rails 3.2 → 4.0 (Strong Parameters)
+- `version-guides/upgrade-4.0-to-4.1.md` - Rails 4.0 → 4.1 (Spring, secrets.yml, enums)
+- `version-guides/upgrade-4.1-to-4.2.md` - Rails 4.1 → 4.2 (ActiveJob, Web Console)
```
JuanVqz, line 208: Not strictly part of #20 — the `upgrade-4.1-to-4.2.md` file already existed but was missing from this listing. Added it here since I was editing the adjacent lines. Happy to drop this one-liner if you prefer the PR stay strictly scoped to 4.0 → 4.1.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
+version: "4.1"
+description: "Breaking change patterns for Rails 4.0 → 4.1 upgrade"
+
+breaking_changes:
+  high_priority:
+    - name: "Dynamic finders (find_all_by / find_last_by / scoped_by)"
+      pattern: "\\.(find_all_by_|find_last_by_|scoped_by_)[a-zA-Z_0-9]+"
```
etagwerker: Would it be worth adding `find_by_` (and `find_by_*!`) to this alternation? `activerecord-deprecated_finders` also supplied the single-record `find_by_foo(val)` / `find_by_foo!(val)`, which are the most common dynamic-finder forms in 4.0-era code and also break in 4.1 once the gem dep is gone. The new hash-form `find_by(col: val)` would need an exclude to avoid false positives, happy to sketch one if useful.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
...
+      exclude: "\\.references\\(|\\.eager_load\\("
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 4.1 removed the string-parsing heuristic that auto-joined tables referenced inside a where(\"other_table.col = ?\") string after includes(...)"
+      fix: "Replace with explicit joins(:assoc) (no eager load), eager_load(:assoc) (eager loaded), or includes(:assoc).where(...).references(:assoc)"
+      variable_name: "IMPLICIT_JOIN_REFERENCES"
+
+    - name: "store_accessor on json/hstore column (PG key semantics)"
+      pattern: "^\\s*store_accessor\\s+:"
```
etagwerker, line 48: Is this pattern actually detecting the bug? The explanation on line 52 notes that `store_accessor`-generated readers/writers are *unaffected*, so the break is on direct symbol-keyed access like `profile.preferences[:theme]`. As written, this fires on every model that uses `store_accessor` (a proxy for "review models using PG json/hstore"), not on the failing pattern itself. Would it be cleaner to reframe the explanation as a manual-audit signal, or try to pattern-match direct `[:symbol]` lookups into known json/hstore columns?

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
...
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+        - "config/"
+      explanation: "Rails 4.1 no longer depends on MultiJSON. Apps that reference it directly will hit NameError once Rails stops pulling it in transitively. Gemfile is intentionally excluded — `gem 'multi_json'` IS the fix, not a breaking-change signal"
+      fix: "Either keep gem 'multi_json' in the Gemfile, or migrate callers to core JSON: obj.to_json and JSON.parse(str). Do NOT blindly replace with JSON.dump/JSON.load — those are unsafe for arbitrary objects"
+      variable_name: "MULTI_JSON"
+
+    - name: "default_scope in models"
```
etagwerker: Should this move to 🔴 HIGH? Per the CLAUDE.md rubric, "silently wrong behavior with production impact" is HIGH, and a reporting/admin query returning zero rows in 4.1 where 4.0 returned data matches that failure mode exactly. Blast radius is narrow (models with a `default_scope` AND a contradicting scope), but when it lands it's silent from the user's perspective. WDYT?

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
...
+      fix: "If no match: add `Rails.application.config.action_dispatch.cookies_serializer = :hybrid` in a new config/initializers/cookies_serializer.rb. :hybrid reads legacy Marshal cookies and writes new JSON ones"
+      variable_name: "COOKIES_SERIALIZER"
+
+    - name: "I18n enforce_available_locales config"
+      pattern: "enforce_available_locales|available_locales"
+      exclude: ""
+      search_paths:
+        - "config/"
+      explanation: "Rails 4.1 defaults enforce_available_locales to true. Any locale not in I18n.available_locales raises I18n::InvalidLocale. Presence check — ABSENCE of both settings is the signal to declare available_locales or disable enforcement"
+      fix: "If neither setting is present: declare config.i18n.available_locales = [...] in config/application.rb. Disabling enforcement (config.i18n.enforce_available_locales = false) is an escape hatch; the default exists for a security reason"
```
etagwerker: Would you consider leading the fix with "audit locale-accepting request paths" before the two config options? The escape hatch (`enforce_available_locales = false`) is already listed last with a caveat, which is good. But naming the actual action (validate inbound locale params) first might save readers from disabling the guardrail as a reflex.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
...
+
+    - name: "Legacy secret_key_base / secret_token reads"
+      pattern: "Rails\\.application\\.config\\.secret_(key_base|token)"
+      exclude: ""
+      search_paths:
+        - "config/"
+        - "app/"
+      explanation: "Matches callers of the old Rails.application.config.secret_key_base / secret_token API. Rails 4.1 introduces config/secrets.yml exposed via Rails.application.secrets. NOTE: this is a migration-work signal based on hits; absence does NOT confirm secrets.yml is configured — grep cannot detect absence of a file"
+      fix: "Create config/secrets.yml with secret_key_base per environment, then migrate readers to Rails.application.secrets.secret_key_base. Delete the old config/initializers/secret_token.rb"
+      variable_name: "SECRET_KEY_BASE_READS"
```
etagwerker: Should this be renamed to `SECRETS_CONFIG_READS` (or similar)? The regex on line 147 matches both `secret_key_base` and `secret_token`, so the current variable name misrepresents half of the hits.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
```diff
@@ -0,0 +1,244 @@
...
+
+    - name: "render :text usage"
+      pattern: "\\brender\\s+(:text\\s*=>|text:)"
```
etagwerker: Would you tweak this to also catch paren-form `render(text: "x")`? The `\s+` after `render` requires whitespace, so `render(text: …)` slips through. Something like `\brender\s*[(\s]\s*(?:.*,\s*)?(?:text:\s*|:text\s*=>)` would cover both. Minor, but paren-form does show up in older 4.0 code.

rails-upgrade/version-guides/upgrade-4.0-to-4.1.md
```diff
@@ -0,0 +1,695 @@
...
+remove `config.active_record.disable_implicit_join_references` from your config.
+
+See [rails/rails#9712](https://github.com/rails/rails/issues/9712) for background.
+
+---
+
+#### 4. PostgreSQL `json` / `hstore` Columns Return String-Keyed Hashes
+
+**What Changed:**
+In 4.0, PostgreSQL `json` and `hstore` columns (and any `store_accessor` built on top of them) returned a `HashWithIndifferentAccess` — symbol and string access both worked. In 4.1 they return a plain `Hash` with **string keys only**. Symbol access silently returns `nil`.
```
etagwerker: Would it be worth extending this one line to mention PostgreSQL **array** columns? The official 4.1 upgrade guide groups arrays with `json`/`hstore` under the same string-key change, and the fix is identical. One sentence keeps the section tight without adding a whole new entry.

etagwerker (commented):
Strong PR overall. Technically accurate on the hard parts (PG json/hstore key flip, inline-block callback `return`, `default_scope` chaining, cookies serializer, CSRF-on-GET-JS), well-sourced, and style-matches the 4.1 → 4.2 guide. A few questions inline, mostly around pattern precision and one priority calibration. Nothing blocking merge.

JuanVqz:
@etagwerker fixed them all.

Quick question: some of these look like deprecation warnings, for example:
[https://github.com/ombulabs/claude-code_rails-upgrade-skill/blob/49b18b7b47b017bed92a69361aa35b9f5fa57540/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml#L6-L14](https://github.com/ombulabs/claude-code_rails-upgrade-skill/blob/49b18b7b47b017bed92a69361aa35b9f5fa57540/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml#L6-L14)

Right now, we’re handling them in **Step 3: Run Breaking Changes Detection**, which happens after the dual-boot process.

https://github.com/ombulabs/claude-code_rails-upgrade-skill/blob/49b18b7b47b017bed92a69361aa35b9f5fa57540/rails-upgrade/SKILL.md?plain=1#L82-L96

It feels like these should be addressed *before* dual-booting to keep things simpler and reduce noise early on.

[https://github.com/ombulabs/claude-code_rails-upgrade-skill/blob/49b18b7b47b017bed92a69361aa35b9f5fa57540/rails-upgrade/SKILL.md?plain=1#L96](https://github.com/ombulabs/claude-code_rails-upgrade-skill/blob/49b18b7b47b017bed92a69361aa35b9f5fa57540/rails-upgrade/SKILL.md?plain=1#L96)

Am I missing something here, or would it make sense to split these patterns so deprecations are handled earlier?

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. Pattern now includes `find_by_*` (with optional `!` for the bang variant), and the hash-form `find_by(col: val)` is excluded via `\.find_by\s*\(`. Resolves.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. The explanation is reframed as a manual-audit signal: the pattern flags models that use `store_accessor`, and the explanation directs the reader to audit callers indexing the underlying json/hstore column with symbol keys. Resolves.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. Moved to `high_priority`, with the explanation calling out silent zero-row returns. Lines up with the rubric's silently-wrong-with-production-impact rule. Resolves.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. Fix now leads with 'Audit request paths that accept a locale param from user input and validate it against an allow-list,' followed by the `available_locales` declaration and the escape hatch. Reads in the right order. Resolves.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. Renamed to `SECRETS_CONFIG_READS`. Matches the regex's actual scope (both `secret_key_base` and `secret_token`). Resolves.

rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
etagwerker: Addressed in ee59b91. Regex tightened to also catch the paren-form `render(text: ...)`. Resolves.

rails-upgrade/version-guides/upgrade-4.0-to-4.1.md
etagwerker: Addressed in 49b18b7. Section header expanded to 'PostgreSQL `json` / `hstore` / `array` Columns Return String-Keyed Data', and the intro covers all three together with the same fix guidance. Resolves.

etagwerker:
@JuanVqz  You're not missing anything. That's a good observation. A lot of the HIGH patterns here have unconditional fixes that work on both 4.0 and 4.1 (e.g., where(col: val), joins(...), false if invalid?). 

They don't need NextRails.next? branching, which echoes invariant #7 in the dual-boot skill: only branch for genuine breaking changes; migrate deprecations directly.

I'd say this is a workflow restructure question (when in the SKILL.md flow does detection run, and is there a "pre-dual-boot deprecation cleanup" phase) rather than a patterns-file question. 

The YAML is correct either way. Want to file a separate issue so we can scope and discuss the workflow change without holding up this PR? I think it's worth doing — cleaner dual-boot diffs would be a real win.

PR #38, opened 2026-04-22
Add Rails 5.0 → 5.1 detection patterns

### LINKED ISSUE #14: Add detection patterns for Rails 5.0 → 5.1 upgrade
Detection patterns exist for Rails 6.0+ upgrades but are missing for older upgrade paths.

The `version-guides/upgrade-5.0-to-5.1.md` guide exists but there is no corresponding `detection-scripts/patterns/rails-51-patterns.yml` file.

## Acceptance criteria

- [x] Create `rails-51-patterns.yml` covering all breaking changes from the version guide
- [x] Organize patterns into `high_priority` and `medium_priority` sections
- [x] Include `dependencies` section for any bridge/compatibility gems
- [x] Verify YAML parses without errors
- [x] Verify naming convention matches `rails-{VERSION}-patterns.yml` for auto-discovery by `direct-detection-workflow.md`

In progress: #38

JuanVqz:
## Summary

Closes #14. Consolidates three related changes for the 5.0 ↔ 5.1 upgrade coverage:

1. **Detection patterns for 5.0 → 5.1** (issue #14).
2. **Guide fix: `redirect_to :back` removal wording** (originally #39, folded in).
3. **Deprecation entry for `redirect_to :back` in the prior 4.2 → 5.0 hop** (originally #40, folded in).

Cross-checked against three sources before drafting:
- `rails-upgrade/version-guides/upgrade-5.0-to-5.1.md` (in-repo guide)
- OmbuLabs ebook chapter `from-5-0-to-5-1.md`
- Official [Upgrading from Rails 5.0 to Rails 5.1](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-0-to-rails-5-1)

---

## Part 1 — detection patterns for 5.0 → 5.1 (issue #14)

Adds `rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml`. 14 patterns across HIGH/MEDIUM plus a `dependencies:` section.

**🔴 HIGH (8)**
- `render :text` removed
- `render :nothing` removed
- `redirect_to :back` removed
- `*_filter` controller callbacks removed in favor of `*_action`
- String `if:` / `unless:` conditions on filters (plus `prepend_*` / `append_*` variants and legacy `_filter` suffix)
- `ActiveRecord::Relation#uniq` removed (use `#distinct`). Search scoped to `app/models/` + `app/controllers/` to cut false positives.
- `use_transactional_fixtures` removed (renamed to `use_transactional_tests` in 5.0)
- Controller-test positional arguments dropped (matches both `{ hash }` and `key: value` forms)

**🟡 MEDIUM (6)**
- `render :body` no longer applies default layout
- `raise_in_transactional_callbacks` dead config (was deprecated in 5.0, removed in 5.1)
- `config.serve_static_files` → `config.public_file_server.enabled`
- `config.static_cache_control` → `config.public_file_server.headers` (value shape changes to a hash)
- `application.secrets` nested keys are symbols now (string access returns nil)
- Top-level `HashWithIndifferentAccess` soft-deprecated (use `ActiveSupport::HashWithIndifferentAccess`)

**Dependencies:** `webpacker` (new 5.1 default, optional) and `rails-controller-testing` (`check: true`, carryover from the 5.0 extraction for `assigns()` / `assert_template`).

### Sources beyond the in-repo 5.0 → 5.1 guide

The in-repo guide had 7 entries. These additional items are breaking changes per ebook and/or official upgrade guide, and are included in the YAML so the detection does not under-report:
- `*_filter` → `*_action` (ebook §4.2)
- String filter conditions must be symbols (ebook §4.2)
- `ActiveRecord::Relation#uniq` removed (ebook §4.1)
- `use_transactional_fixtures` renamed (ebook §4.1)
- `config.serve_static_files` / `config.static_cache_control` renames (ebook §3)
- `application.secrets` nested-symbol access (official guide §10.2)
- Top-level `HashWithIndifferentAccess` soft-deprecation (official guide §10.1)

---

## Part 2 — 5.0 → 5.1 guide fix (was PR #39)

The in-repo `upgrade-5.0-to-5.1.md` listed `redirect_to :back` under 🟡 MEDIUM and described it as "deprecated". It was deprecated in Rails 5.0 but **removed** in Rails 5.1 — callers raise at runtime.

- Moved the entry from 🟡 MEDIUM to 🔴 HIGH.
- Reworded What Changed: "was deprecated in Rails 5.0 and **removed** in Rails 5.1. Callers raise at runtime."
- Added a note about `fallback_location:` being required when `HTTP_REFERER` may be missing.
- Split the matching Common Issues entry into two distinct cases so the Cause lines are accurate:
  1. `redirect_to :back` raising because it was removed in 5.1.
  2. `redirect_back` itself raising `RedirectBackError` when the referer is missing and no `fallback_location:` was given — a separate, ongoing hazard of the new API.

---

## Part 3 — 4.2 → 5.0 deprecation entry (was PR #40)

Closes the blind spot where a user upgrades to 5.0, never sees `redirect_to :back` flagged by the skill, then hits runtime errors on the 5.1 hop.

- `rails-upgrade/version-guides/upgrade-4.2-to-5.0.md` → new #10 entry under 🟡 MEDIUM with What Changed / detection / BEFORE-AFTER and a forward note pointing at the 5.1 removal.
- `rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml` → matching MEDIUM pattern `REDIRECT_TO_BACK_DEPRECATED` scoped to `app/controllers/`.

### Convention applied across both hops

- Deprecation-introduction hop → 🟡 MEDIUM, framing: "deprecated — fix now before the next hop."
- Removal hop → 🔴 HIGH, framing: "removed — raises at runtime."

---

## Acceptance criteria (issue #14)

- [x] Create `rails-51-patterns.yml` covering all breaking changes from the version guide
- [x] Organize patterns into `high_priority` and `medium_priority` sections
- [x] Include `dependencies:` section for any bridge/compatibility gems
- [x] Verify YAML parses without errors
- [x] Verify naming convention matches `rails-{VERSION}-patterns.yml` for auto-discovery by `direct-detection-workflow.md`

## Test plan

- [x] `ruby -e "require 'yaml'; YAML.safe_load_file('rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml')"` succeeds
- [x] `ruby -e "require 'yaml'; YAML.safe_load_file('rails-upgrade/detection-scripts/patterns/rails-50-patterns.yml')"` succeeds
- [x] All regexes compile under Ruby's Oniguruma
- [ ] Manual smoke: run the detection on a Rails 5.0 app and confirm the HIGH patterns (`*_filter`, `redirect_to :back`, `render :text`, controller-test positional args) fire where expected
- [ ] Manual smoke: run the detection on a Rails 4.2 app and confirm `REDIRECT_TO_BACK_DEPRECATED` fires on `redirect_to :back` call sites

### DIFF
--- a/rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
@@ -1,0 +1,155 @@
+version: "5.1"
+description: "Breaking change patterns for Rails 5.0 → 5.1 upgrade"
+
+breaking_changes:
+  high_priority:
+    - name: "render :text removed"
+      pattern: "\\brender\\s+(:text\\s*=>|text:)"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 removes `render text: ...`. Calls raise ArgumentError at runtime"
+      fix: "Use `render plain: 'content'` for text/plain, or `render html: '<p>..'.html_safe` for explicit HTML"
+      variable_name: "RENDER_TEXT"
+
+    - name: "render :nothing removed"
+      pattern: "\\brender\\s+(:nothing\\s*=>|nothing:)"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 removes `render nothing: true`. Calls raise ArgumentError at runtime"
+      fix: "Replace with `head :ok` (200 + empty body) or `head :no_content` (204)"
+      variable_name: "RENDER_NOTHING"
+
+    - name: "redirect_to :back removed"
+      pattern: "redirect_to\\s+:back\\b"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 removes `redirect_to :back` outright (it was deprecated in 5.0). Callers raise at runtime"
+      fix: "Replace with `redirect_back(fallback_location: root_path)` and pass any flash/notice options as keyword arguments"
+      variable_name: "REDIRECT_TO_BACK"
+
+    - name: "*_filter callback methods removed"
+      pattern: "\\b(before|after|around|skip_before|skip_after|skip_around|prepend_before|prepend_after|prepend_around|append_before|append_after|append_around)_filter\\b"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+        - "app/"
+        - "lib/"
+      explanation: "Rails 5.1 removes all `*_filter` controller callbacks (deprecated in 4.x, still warning in 5.0). Use `*_action` variants"
+      fix: "Rename: before_filter → before_action, skip_before_filter → skip_before_action, etc. Mechanical rename — behavior is identical"
+      variable_name: "FILTER_METHODS"
+
+    - name: "String conditions in controller filters"
+      pattern: "(before|after|around|skip_before|skip_after|skip_around)_action\\b[^\\n]*\\b(if|unless):\\s*['\"]"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 no longer accepts string values for `if:` / `unless:` conditions on filters. They must be symbols (method names) or lambdas"
+      fix: "Replace the string with a symbol: `if: :has_project_guest_id` instead of `if: 'has_project_guest_id'`. For inline expressions, use a lambda: `if: -> { ... }`"
+      variable_name: "FILTER_STRING_CONDITIONS"
+
+    - name: "ActiveRecord::Relation#uniq removed"
+      pattern: "\\.uniq(\\b|\\()"
+      exclude: "\\.(to_a|map|pluck|select|reject|each|values|keys|compact)[^\\n]*\\.uniq"
+      search_paths:
+        - "app/"
+        - "lib/"
+      explanation: "Rails 5.1 removes ActiveRecord::Relation#uniq (was deprecated in 5.0). Use #distinct. NOTE: pattern will false-positive on Array#uniq — the exclude filters common array cases but verify receivers are Relations"
+      fix: "Replace .uniq / .uniq! on ActiveRecord relations with .distinct. Plain Array#uniq is unaffected — keep those as-is"
+      variable_name: "RELATION_UNIQ"
+
+    - name: "use_transactional_fixtures removed"
+      pattern: "use_transactional_fixtures"
+      exclude: ""
+      search_paths:
+        - "test/"
+        - "spec/"
+      explanation: "Rails 5.1 removes `use_transactional_fixtures` (renamed to `use_transactional_tests` in Rails 5.0)"
+      fix: "Rename the setting to `self.use_transactional_tests = true` in your test base class / test_helper.rb"
+      variable_name: "USE_TRANSACTIONAL_FIXTURES"
+
+    - name: "Controller test positional arguments"
+      pattern: "^\\s*(get|post|put|patch|delete|head)\\s+[:'\"][^,\\n]+,\\s*\\{"
+      exclude: "params:|session:|flash:|headers:|body:|xhr:|format:"
+      search_paths:
+        - "test/controllers/"
+        - "test/functional/"
+        - "spec/controllers/"
+        - "spec/requests/"
+      explanation: "Rails 5.1 drops support for positional arguments in controller test helpers. All parameters must be keyword-argument form"
+      fix: "Wrap params/session/flash in keyword args: `get :show, params: { id: 1 }, session: { user_id: 1 }, flash: { notice: 'hi' }`"
+      variable_name: "CONTROLLER_TEST_POSITIONAL"
+
+  medium_priority:
+    - name: "render :body default layout change"
+      pattern: "\\brender\\s+(:body\\s*=>|body:)"
+      exclude: "layout:"
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 changes `render body:` to NOT apply the default layout. If your 5.0 code relied on the layout being rendered, output will differ silently"
+      fix: "Pass `layout: true` (or a layout name) explicitly if you need the layout. If you wanted raw body output, no change needed"
+      variable_name: "RENDER_BODY_LAYOUT"
+
+    - name: "raise_in_transactional_callbacks config (dead setting)"
+      pattern: "raise_in_transactional_callbacks"
+      exclude: ""
+      search_paths:
+        - "config/"
+      explanation: "Rails 5.1 removes the raise_in_transactional_callbacks config option. The 5.0+ behavior (always raise) is now the only behavior"
+      fix: "Remove `config.active_record.raise_in_transactional_callbacks = true` from config/application.rb or environment configs"
+      variable_name: "RAISE_IN_TRANSACTIONAL_CALLBACKS"
+
+    - name: "serve_static_files config renamed"
+      pattern: "config\\.serve_static_files"
+      exclude: ""
+      search_paths:
+        - "config/"
+      explanation: "Rails 5.1 renames `config.serve_static_files` to `config.public_file_server.enabled`"
+      fix: "Rename: `config.public_file_server.enabled = <value>` (or `= ENV['RAILS_SERVE_STATIC_FILES'].present?` for the production default)"
+      variable_name: "SERVE_STATIC_FILES_RENAME"
+
+    - name: "static_cache_control config renamed"
+      pattern: "config\\.static_cache_control"
+      exclude: ""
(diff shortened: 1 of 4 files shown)


rails-upgrade/version-guides/upgrade-5.0-to-5.1.md
JuanVqz, line 1: There is more context on this change here (#39), but basically this is not a deprecation warning happening between Rails 5.0 and 5.1; the deprecation warning was from Rails 4.2 to 5.0. at this point this is raising an error if not addressed.

Why do we keep it if it is a deprecation for a previous upgrade hop? well, because it appears in both upgrade guides, rails 4.2 to 5.0 and 5.0 to 5.1, but it changes a bit from a deprecation warning to asolve this or it is going to break.

rails-upgrade/version-guides/upgrade-4.2-to-5.0.md
JuanVqz, line 1: Just adding a missing deprecation warning in this bump

etagwerker (commented):
Solid consolidation. The three-part framing (5.1 patterns + 5.0 → 5.1 guide fix + back-filled deprecation entry on the prior hop) is exactly the right move, and the `redirect_to :back` MEDIUM → HIGH correction between 5.0 and 5.1 closes a real trap. YAML parses clean (8 HIGH, 6 MEDIUM, 2 deps). A few technical questions inline, mostly around whether a couple of entries are actually 5.1-specific and whether the HIGH/MEDIUM calls match removal vs deprecation. Nothing blocking.

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
```diff
@@ -0,0 +1,155 @@
...
+      explanation: "Rails 5.1 removes `redirect_to :back` outright (it was deprecated in 5.0). Callers raise at runtime"
+      fix: "Replace with `redirect_back(fallback_location: root_path)` and pass any flash/notice options as keyword arguments"
+      variable_name: "REDIRECT_TO_BACK"
+
+    - name: "*_filter callback methods removed"
+      pattern: "\\b(before|after|around|skip_before|skip_after|skip_around|prepend_before|prepend_after|prepend_around|append_before|append_after|append_around)_filter\\b"
+      exclude: ""
+      search_paths:
+        - "app/controllers/"
+        - "app/"
```
etagwerker: Would you consider narrowing `search_paths` to just `app/controllers/` and `lib/`? `app/` is a superset of `app/controllers/`, so the current list is redundant, and it'll also scan `app/models/`, `app/helpers/`, `app/views/`, etc., where `*_filter` call sites are extremely unlikely.

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
```diff
@@ -0,0 +1,155 @@
...
+
+    - name: "Controller test positional arguments"
+      pattern: "^\\s*(get|post|put|patch|delete|head)\\s+[:'\"][^,\\n]+,\\s*(\\{|[a-z_]+:)"
+      exclude: "params:|session:|flash:|headers:|body:|xhr:|format:"
+      search_paths:
+        - "test/controllers/"
+        - "test/functional/"
+        - "spec/controllers/"
+        - "spec/requests/"
+      explanation: "Rails 5.1 drops support for positional arguments in controller test helpers. All parameters must be keyword-argument form"
```
etagwerker, line 80: The wording "drops support" is a bit ambiguous. Were positional test args actually *removed* in 5.1 (raises at runtime), or still *deprecated-with-warning* (final removal came later)? My recollection is closer to the latter, and worth double-checking against the 5.1 ActionPack CHANGELOG. If it's still deprecation-only in 5.1, the CLAUDE.md rubric would land this at 🟡 MEDIUM rather than 🔴 HIGH.

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
```diff
@@ -0,0 +1,155 @@
...
+      fix: "Wrap params/session/flash in keyword args: `get :show, params: { id: 1 }, session: { user_id: 1 }, flash: { notice: 'hi' }`"
+      variable_name: "CONTROLLER_TEST_POSITIONAL"
+
+  medium_priority:
+    - name: "render :body default layout change"
+      pattern: "\\brender\\s+(:body\\s*=>|body:)"
+      exclude: "layout:"
+      search_paths:
+        - "app/controllers/"
+      explanation: "Rails 5.1 changes `render body:` to NOT apply the default layout. If your 5.0 code relied on the layout being rendered, output will differ silently"
```
etagwerker: Could you point me at the source for this one? I couldn't find an explicit "`render body:` stops applying default layout" entry in the 5.1 release notes or ActionPack CHANGELOG. If the behavior has been the same since 5.0 (i.e., `render body:` never applied the default layout), the entry may mislead readers into thinking a 5.0 → 5.1 hop introduces silent output differences that it actually doesn't.

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
```diff
@@ -0,0 +1,155 @@
...
+      pattern: "config\\.static_cache_control"
+      exclude: ""
+      search_paths:
+        - "config/"
+      explanation: "Rails 5.1 renames `config.static_cache_control` to `config.public_file_server.headers` (and the value is now a hash, not a string)"
+      fix: "Rename and restructure: `config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }`"
+      variable_name: "STATIC_CACHE_CONTROL_RENAME"
+
+    - name: "application.secrets nested string-key access"
+      pattern: "Rails\\.application\\.secrets\\[[^\\]]+\\]\\[['\"]"
```
etagwerker, line 113: The regex catches `Rails.application.secrets[:key]["nested"]` bracket-chain access, but the canonical style in Rails docs is usually `Rails.application.secrets.smtp_settings["address"]` (method accessor + bracket). Should we add a second pattern for that form so we don't miss what's likely the more common call site?

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
```diff
@@ -0,0 +1,155 @@
...
+      variable_name: "SECRETS_STRING_KEYS"
+
+    - name: "Top-level HashWithIndifferentAccess constant"
+      pattern: "(?<!ActiveSupport::)\\bHashWithIndifferentAccess\\b"
+      exclude: ""
+      search_paths:
+        - "app/"
+        - "lib/"
+        - "config/"
+      explanation: "Rails 5.1 soft-deprecates the top-level HashWithIndifferentAccess constant in favor of ActiveSupport::HashWithIndifferentAccess. No warning yet, but the constant will be removed in a future version"
```
etagwerker: The explanation says "no warning yet." If Rails 5.1 doesn't actually emit a deprecation for the top-level `HashWithIndifferentAccess` constant, should this entry live in whichever release surfaces the warning (5.2? 6.x?) rather than 5.1? Or, if the intent is purely forward-looking cleanup during a 5.1 upgrade, would 🟢 LOW be a better fit than 🟡 MEDIUM given there's nothing yet forcing the user's hand?

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
JuanVqz: Verified against [Rails 5.1 ActionPack CHANGELOG](https://github.com/rails/rails/blob/5-1-stable/actionpack/CHANGELOG.md?plain=1#L368-L369). The entries say:

```
Remove deprecated support to non-keyword arguments in ActionDispatch::IntegrationTest#process, #get, #post, #patch, #put, #delete, and #head.
```

And same for ActionController::TestCase.

Positional args were deprecated in 5.0 or earlier, but Rails 5.1 actually **removes** support (raises at runtime). Current wording "drops support" is accurate, and HIGH priority is justified — it's a breaking change that breaks tests at runtime, not a soft deprecation. ✓

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
JuanVqz: Added a second pattern for this, to don't make it complicated and cover both

rails-upgrade/detection-scripts/patterns/rails-51-patterns.yml
JuanVqz: I just removed it

JuanVqz:
@etagwerker thanks for the feedback, I just address them all, ready to be reviewed again.

etagwerker (approved):

PR #47, opened 2026-04-29
Add upgrade-cleanup plugin

### LINKED ISSUE #1: Improve skill and add a clean up task before jumping to the next jump
Henrique mentioned that it would be useful to clean up all the calls to `NextRails.next?` before jumping to the next version jump.

JuanVqz:
## Summary

Adds an `upgrade-cleanup` sibling plugin for finishing a Rails upgrade campaign and removing dual-boot scaffolding. Scoped tightly to **scaffolding removal**, not a kitchen-sink "finish the upgrade" pass.

**In scope:**

- **Phase 0, Pre-flight.** Detect run environment (Docker vs local) and ask the user when ambiguous. Smoke-check `bundle` and `bin/rails runner` on both sides before tearing down `Gemfile.next`. Stop if the next side has rotted (cleanup is premature when the upgrade itself isn't actually working).
- **Phase 1, Dual-boot removal.** Drop `NextRails.next?` / `NextRails.current?` branches from app code, strip `next?` Gemfile method + `if next?` conditionals, sweep for `deprecation_tracker` residue *before* removing `next_rails` (the gem ships `DeprecationTracker`, leftover `require`s break test boot), remove `next_rails`, swap lockfiles, run `bundle install` (incremental, not from-scratch), update CI to drop the dual-boot job. Includes a sanity check for the case where `Gemfile.next.lock` was never regenerated and is byte-identical to the old lock.
- **Phase 2, Old-version code retirement.** Monkey-patches, stale gem pins, conditional Gemfile groups, `docker-compose.yml` / `compose.yaml` sister services (`web-next`, `worker-next` with `BUNDLE_GEMFILE: Gemfile.next`), doc drift (`README`, `bin/setup`, `.tool-versions`, `Dockerfile`).
- **Phase 3, Housekeeping.** CI matrix entry, `Dockerfile` / `.ruby-version` / `.tool-versions` alignment for Ruby bumps.
- **Phase 4, Final verification.** Local or CI, with explicit fallback to CI when the local environment can't run tests.
- **Phase 5, Commit and PR.** Suggested commit messages, single-purpose PR.

**Out of scope (deliberately):**

- `load_defaults` alignment. Already owned by the rails-upgrade skill via the `rails-load-defaults` dependency. Cleanup does not duplicate it.
- Deprecation triage. Surfaced as a "Next Steps" pointer at the end of the workflow, but the actual triage belongs to the rails-upgrade skill's next-hop workflow.
- Migration class suffix and `db/schema.rb` regen. Upgrade artifacts, not cleanup.

**Design choices worth noting:**

- Lives as a sibling plugin (`upgrade-cleanup/.claude-plugin/plugin.json` + skill + workflow), invoked via auto-trigger phrases ("finish the upgrade", "clean up dual-boot"). The earlier `/upgrade-cleanup` slash command and the duplicate workflow inside `rails-upgrade/workflows/` were dropped to avoid drift between two sources of truth.
- Phase 1 owns the scaffolding-removal step list directly rather than delegating to the `dual-boot` skill, so the workflow is self-contained when the dual-boot plugin isn't installed. SKILL.md notes this skill wins on conflict.

**Validated against a real cleanup:**

Ran the workflow end-to-end on `ombulabs/points` (`fake/add-more-dual-boot-conditionals` branch, keep-current direction, abandoning a Rails 7.1 → 7.2 attempt). 204/204 specs green after cleanup. Two real-world gaps surfaced and were folded back into the workflow:

- `next_rails` ships `DeprecationTracker`, often wired into `spec/rails_helper.rb` via `require "deprecation_tracker"` + `DeprecationTracker.track_rspec(...)`. Removing the gem first breaks test boot with `LoadError`; the workflow now sweeps for the residue *before* the gem-removal step.
- `docker-compose.yml` / `compose.yaml` commonly carry sister services (`web-next`, `worker-next`) with `BUNDLE_GEMFILE: Gemfile.next` via YAML anchors. Phase 2 previously only mentioned `Dockerfile`; it now has a dedicated step covering both compose files.

Closes #1.

## Test plan

- [x] CI green on the branch
- [x] Manual: invoke via "finish the upgrade" / "clean up dual-boot" on a real dual-boot repo (validated on `ombulabs/points`, keep-current direction, 204/204 specs green)
- [x] Manual: confirm `bundle install` drops `next_rails` from `Gemfile.lock` without re-resolving `rails` (verified via `git diff Gemfile.lock`, only `next_rails` + `colorize` removed)
- [ ] Manual: invoke on a different repo and confirm Phase 0 fails fast when `Gemfile.next` is broken
- [ ] Manual: confirm Phase 4 falls back to CI as the validating environment when the local shell can't reach a database


rails-upgrade/SKILL.md
```diff
@@ -108,12 +108,20 @@ When proposing code fixes that must work with both the current and target Rails
...
 
-### Step 6: Align load_defaults to New Version (FINAL STEP)
+### Step 6: Align load_defaults to New Version
 - **DELEGATE** to the `rails-load-defaults` skill for detection and incremental update
 - That skill handles tiered, per-config updates with test runs between each change
-- This is done AFTER the Rails version upgrade is complete, as the last step
+- This is done AFTER the Rails version upgrade is complete
 - **DEPENDENCY:** Requires the [rails-load-defaults skill](https://github.com/ombulabs/claude-code_rails-load-defaults-skill)
 
+### Step 7: Suggest Cleanup (USER-TRIGGERED)
```
JuanVqz, line 117: I think we can suggest the users to clean the dual-boot when they decide not on each upgrade. Because if they want to do multiple jumps, they don't need to clean up each time if they will install all again.

upgrade-cleanup/upgrade-cleanup/workflows/upgrade-cleanup-workflow.md
```diff
@@ -0,0 +1,177 @@
...
+- **Next-side rails runner fails but bundle succeeds:** boot regression on the new version. Tell the user; don't tear down the rollback path until it's fixed.
+- **Current-side bundle fails:** environment problem unrelated to the upgrade. Resolve before continuing — you don't want a half-broken environment during cleanup.
+
+If the user explicitly says *"skip the pre-flight, I know it works"*, record their override and continue. Their call.
+
+---
+
+## Phase 1: Remove `NextRails.next?` / `NextRails.current?` Branches and Dual-Boot Scaffolding
+
+This skill owns these steps. The `dual-boot` skill's `workflows/cleanup-workflow.md` is older background reference.
```
JuanVqz: I wonder if this is fine or we should move that cleanup-workflow to this plugin.

upgrade-cleanup/upgrade-cleanup/workflows/upgrade-cleanup-workflow.md
```diff
@@ -0,0 +1,177 @@
...
+
+---
+
+## Phase 5: Final Verification
+
+Before declaring cleanup done:
+
+- [ ] Test suite passes on the upgraded version (no dual-boot, single Gemfile)
+- [ ] CI is green on the cleanup branch
+- [ ] `grep -rE "NextRails\.(next|current)\?" . --include="*.rb"` returns nothing
```
JuanVqz, line 167: What about erb or haml files, should we check those as well?

upgrade-cleanup/upgrade-cleanup/workflows/upgrade-cleanup-workflow.md
```diff
@@ -0,0 +1,165 @@
...
+4. Remove all `if next?` / `else` conditionals from the `Gemfile` (keep new-version gems).
+5. Remove the `next_rails` gem from the `Gemfile`.
+6. Replace `Gemfile.lock` with `Gemfile.next.lock` (`mv Gemfile.next.lock Gemfile.lock`), then delete `Gemfile.next`. This is the standard step. It preserves the exact gem versions tested during the upgrade. Do not run `bundle update` or delete the lockfile to re-resolve from scratch, that risks drift on versions you already validated.
+7. Run `bundle install` (NOT `bundle update`). The swapped lockfile still pins `next_rails` and any other dual-boot-only gems because `Gemfile.next` listed them. `bundle install` is incremental: it removes references to gems no longer in the `Gemfile` without re-resolving the rest. `rails` and friends stay pinned.
+8. Run the project's test suite (detect the runner from the `Gemfile`: `rspec-rails` means `bundle exec rspec`, otherwise `bundle exec rake test` or `bin/rails test`).
+9. Update CI to drop the dual-boot job/matrix entry.
```
JuanVqz: added those 9 steps as fallback if the dual-boot skill does not exist

rails-upgrade/SKILL.md
```diff
@@ -359,6 +375,17 @@ Claude runs detection directly using tools - NO script generation needed
 4. Consolidates into config/application.rb when done
 ```
 
+### Step 8: Mention Cleanup (USER-TRIGGERED)
```
JuanVqz, line 378: I think it is better to trigger this manually rather than get the agent to do it automatically, just in case users want to go for the next jump. but open to suggestions.

etagwerker (approved):
@JuanVqz I think this change looks good and I only have one comment: It's a bit ambiguous when it talks about "cleaning up the if/else branches." 

Maybe we should instruct Claude to confirm whether the user wants to drop the "current version" branch or the "next version" branch. 

I can see it both ways: 

1. Clean up because we gave up on the upgrade for now (keep current branch) 
2. Clean up because we are confident the next branch works in development, test, production, staging (keep the next branch)

JuanVqz:
> @JuanVqz I think this change looks good and I only have one comment: It's a bit ambiguous when it talks about "cleaning up the if/else branches."
> 
> Maybe we should instruct Claude to confirm whether the user wants to drop the "current version" branch or the "next version" branch.
> 
> I can see it both ways:
> 
> 1. Clean up because we gave up on the upgrade for now (keep current branch)
> 2. Clean up because we are confident the next branch works in development, test, production, staging (keep the next branch)

@etagwerker Good call out to cover both scenarios.

Question:
1. There is a cleanup workflow at dual-boot skill, should we keep it or move it over [here](https://github.com/ombulabs/claude-code_dual-boot-skill/blob/main/dual-boot/workflows/cleanup-workflow.md)?


Done, now the skill ask you which one do you want to keep current or next rails version? 

Tested it and it worked 👍

Issue #53, opened 2026-05-04
Add `kind:` field to detection patterns; rename `breaking_changes:` → `upgrade_findings:`


JuanVqz:
## Problem

The top-level `breaking_changes:` key in `rails-upgrade/detection-scripts/patterns/rails-*-patterns.yml` is a misnomer. Across the 12 version files (4.0 → 8.1, 189 patterns total), the majority of entries are **deprecation warnings**, **migration targets**, or **opt-in features** — not strict breaking changes. Calling them all "breaking changes" loses signal: the agent (and the user reading detection output) cannot distinguish "fix before bump or app won't boot" from "warning today, fix when convenient".

This was surfaced by Copilot review feedback on #43 ([discussion](https://github.com/ombulabs/claude-code_rails-upgrade-skill/pull/43#discussion_r3142610480)) and confirmed during code review of the 5.1 → 5.2 patterns: 12 of 14 entries in that file are deprecations or migrations. The same pattern holds across the rest of the version files.

### 1. Add a `kind:` field per pattern

| Value | Meaning |
|---|---|
| `breaking` | Raises / removed / won't boot. Must fix to upgrade. |
| `deprecation` | Warning at this hop; removal scheduled later. |
| `migration` | Works today, recommended migration target (e.g., `secrets.yml` → `credentials.yml.enc`). |
| `optional` | Opt-in feature or improvement (e.g., bootsnap). |

### 2. Rename the top-level key

`breaking_changes:` → `upgrade_findings:`

More accurate (covers breaks + deprecations + migrations + optional). Reads well: `upgrade_findings.high_priority`.

`priority` (HIGH / MEDIUM / LOW) stays — it expresses **urgency**. `kind:` expresses **what the change is**. The two are orthogonal: a deprecation can still be HIGH priority if it's silently wrong (e.g., `DIRTY_TRACKING_AFTER_SAVE`).

## Phasing

Each step is a separate PR for safe, reviewable rollout:

- [x] #54 — Schema: extend `bin/validate-patterns` to accept `kind:` field (optional during rollout, required after validator-tighten) #70 
- [x] #55 — Classify `kind:` for `rails-40-patterns.yml` (32 patterns, 3.2 → 4.0) #73
- [x] #56 — Classify `kind:` for `rails-41-patterns.yml` (22 patterns, 4.0 → 4.1) #72 
[...]

PR #72, opened 2026-05-04
Classify kind: for rails-41-patterns.yml (#56)

### LINKED ISSUE #56: Classify kind: for rails-41-patterns.yml (4.0 → 4.1)
**Parent:** #53

## Scope

Classify the `kind:` for every pattern in `rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml` (22 patterns total — Rails 4.0 → 4.1 hop).

## Verification (per entry)

- [ ] `kind:` value cross-checked against the **official Rails upgrade guide** for the 4.0 → 4.1 hop
- [ ] Cross-checked against the matching in-repo `rails-upgrade/version-guides/upgrade-4.1-to-*.md`
- [ ] `bin/validate-patterns rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml` passes

## Out of scope

- Reclassifying `priority`

JuanVqz:
## Summary

Classifies the new `kind:` field for every pattern in `rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml` (Rails 4.0 → 4.1 hop, 22 entries). Part of the schema rollout tracked in #53; closes sub-issue #56.

## Classification breakdown

| variable_name | kind | rationale |
|---|---|---|
| DYNAMIC_FINDERS | breaking | `find_all_by_*` etc. removed; raises NoMethodError without bridge gem. |
| FIND_OR_X_BY_DYNAMIC | breaking | Dynamic `find_or_*_by_<col>` forms removed; raises. |
| CALLBACK_RETURN | breaking | `return` inside inline callback block now raises LocalJumpError. |
| IMPLICIT_JOIN_REFERENCES | breaking | String-parsing heuristic removed; queries raise without explicit `references`. |
| PG_JSON_HSTORE_KEYS | breaking | json/hstore now string-keyed; symbol indexing silently returns nil (data-correctness break). |
| DEFAULT_SCOPE | breaking | Default scope now ANDs with subsequent scopes; can silently return zero rows. |
| MULTI_JSON | breaking | Rails no longer pulls in `multi_json`; direct callers raise NameError. |
| RELATION_MUTATORS | breaking | Mutator methods removed from `ActiveRecord::Relation`; raises NoMethodError. |
| JS_CONTROLLER_TEST | breaking | CSRF protection extended to JS GETs; existing tests fail. |
| FLASH_SYMBOL_KEYS | breaking | `flash.to_hash` keys are strings; symbol filters silently no-op (correctness break). |
| COOKIES_SERIALIZER | optional | New JSON serializer is opt-in for upgraded apps; `:hybrid` recommended but not required. |
| I18N_ENFORCE_LOCALES | breaking | `enforce_available_locales` defaults to true; unknown locales raise `I18n::InvalidLocale`. |
| JSON_TIME_PRECISION | migration | New default adds milliseconds; no warning, recommended to set explicitly to match client expectations. |
| SPRING | optional | Ships as new default for new apps; existing apps may add it but need not. |
| SECRETS_CONFIG_READS | migration | Old `secret_token` API still works; `secrets.yml` is the recommended forward path. |
| RENDER_TEXT | migration | `render :text` still works in 4.1; `:plain` / `:html` / `:body` are the recommended replacements ahead of later deprecation. |
| ENCODE_JSON_HOOK | breaking | New encoder no longer calls `encode_json`; custom impls become silently dead (correctness break). |
| BIG_DECIMAL_AS_STRING | migration | Setting is a no-op; cosmetic cleanup, no warning emitted. |
| JSON_GEM_ISOLATION | migration | `JSON.generate`/`dump` no longer honor `as_json`; recommended migration to `to_json` for Rails objects. |
| SET_CALLBACK_AROUND | breaking | Around-callback lambda signature changed; old `&block` form will misbehave. |
| DISABLE_IMPLICIT_JOIN_REFERENCES | deprecation | Setting is a no-op and emits a deprecation warning at runtime. |
| CHECK_PENDING_MIGRATION | optional | Explicit call is harmless but redundant; cleanup only. |

Totals: breaking 13, deprecation 1, migration 5, optional 3.

### Borderline calls

- **JSON_TIME_PRECISION** — could be argued `breaking` (silent serialization change with potential client impact), but Rails itself does not raise or warn, and the explanation frames it as a default to consciously accept or override. Classed as `migration` (recommended path).
- **RENDER_TEXT** — explanation says it "signals :text for future deprecation" but does not state Rails 4.1 emits a warning. Treated as `migration`; will become `deprecation` proper in the 4.1 → 4.2 entry.
- **PG_JSON_HSTORE_KEYS / FLASH_SYMBOL_KEYS / DEFAULT_SCOPE** — silent semantic shifts rather than raises. Classed as `breaking` per the rubric ("silently wrong behavior with production impact").

## Test plan

- [x] `bin/validate-patterns rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml` passes
- [x] `bin/validate-patterns` (all files) passes

## Linked issues

Refs #53
Closes #56

### DIFF
--- a/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-41-patterns.yml
@@ -4,6 +4,7 @@ description: "Breaking change patterns for Rails 4.0 → 4.1 upgrade"
 breaking_changes:
   high_priority:
     - name: "Dynamic finders (find_all_by / find_last_by / scoped_by / find_by)"
+      kind: breaking
       pattern: "\\.(find_all_by_|find_last_by_|scoped_by_|find_by_)[a-zA-Z_0-9]+!?"
       exclude: "\\.find_by\\s*\\("
       search_paths:
@@ -14,6 +15,7 @@ breaking_changes:
       variable_name: "DYNAMIC_FINDERS"
 
     - name: "Dynamic find_or_initialize_by_* / find_or_create_by_*"
+      kind: breaking
       pattern: "\\.(find_or_initialize_by_|find_or_create_by_)[a-zA-Z_0-9]+"
       exclude: ""
       search_paths:
@@ -24,6 +26,7 @@ breaking_changes:
       variable_name: "FIND_OR_X_BY_DYNAMIC"
 
     - name: "return inside inline callback block"
+      kind: breaking
       pattern: "^\\s*(before_|after_|around_)(save|create|update|destroy|validation|commit|rollback|initialize|find|touch)\\b.*\\{[^}]*\\breturn\\b"
       exclude: ""
       search_paths:
@@ -35,6 +38,7 @@ breaking_changes:
       variable_name: "CALLBACK_RETURN"
 
     - name: "Implicit join reference in includes+where string"
+      kind: breaking
       pattern: "\\.includes\\([^)]+\\)[^\\n]*\\.where\\(\\s*['\"][a-z_][a-z_0-9]*\\."
       exclude: "\\.references\\(|\\.eager_load\\("
       search_paths:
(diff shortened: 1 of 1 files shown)


JuanVqz:
Code-reviewed end-to-end with Claude. Ready for human review.

Issue #94, opened 2026-05-05
Thread kind/bucket taxonomy through SKILL.md (follow-up to #69)


JuanVqz:
**Parent / context:** issue #53, follow-up to PR #93 (#69)

## Problem

PR #93 (#69) introduced a `kind`-driven two-bucket sort in `rails-upgrade/workflows/direct-detection-workflow.md`:

- **Fix before bump** — `kind: breaking` and `kind: deprecation`
- **Fix when ready** — `kind: migration` and `kind: optional`

`rails-upgrade/SKILL.md` still uses pre-#53-rollout terminology. The two docs are inconsistent, and a reader landing on SKILL.md first will not see the bucket model that the detection workflow now uses.

## Drift to fix in SKILL.md

| Line | Current | Suggested |
|---|---|---|
| ~301 | *"a report covering breaking changes (with OLD vs NEW code examples…)"* | Mention both buckets — fix-before-bump and fix-when-ready — and that the report should preserve the bucketing from detection |
| ~311 | *"Implement breaking change fixes using `NextRails.next?` for dual-boot code"* | *"Implement fix-before-bump changes (`kind: breaking` and `kind: deprecation`) using `NextRails.next?` for dual-boot code"* |
| ~318 | *"Do not fix deprecations printed by the next version during this hop."* | Disambiguate: this caveat is specifically about `load_defaults`-triggered runtime deprecation warnings that surface after the bump (Step 7's domain), not about `kind: deprecation` detection patterns (which #69 puts in fix-before-bump). |
| ~426 (similar text in Pattern 1 / Common Request Patterns block) | Same "fix breaking changes" phrasing | Same update |

## Acceptance

- [ ] Step 5 (Generate Reports) and Step 6 (Implement Changes) both use the kind/bucket terminology consistently with `direct-detection-workflow.md`
- [ ] Line 318's "Do not fix deprecations printed by the next version" caveat clarified so readers don't think it contradicts fix-before-bump
- [ ] Quality Checklist / Pre-Upgrade sections updated if they still say "breaking changes" generically
- [ ] Common Request Patterns and Trigger Patterns audited for the same drift
- [ ] No re-work of the detection workflow doc — that's #93's territory

## Out of scope

- Any change to YAML pattern files
- Any change to `bin/validate-patterns`
- Any change to `direct-detection-workflow.md`
- The `load_defaults` step (#67-style) — separate concern

## Why now (vs immediately during #69)

[...]

PR #73, opened 2026-05-04
Classify kind: for rails-40-patterns.yml (#55)

JuanVqz:
## Summary

Sub-issue #55 of the issue #53 rollout. Classifies the new `kind:` field for every pattern in `rails-upgrade/detection-scripts/patterns/rails-40-patterns.yml` (32 patterns total — Rails 3.2 → 4.0 hop). `kind:` is placed immediately after `name:` for visual scannability. No other fields touched; `dependencies:` section unchanged.

## Counts

| Kind | Count |
|---|---|
| `breaking` | 25 |
| `deprecation` | 5 |
| `migration` | 2 |
| `optional` | 0 |

### HIGH priority

| Variable | Kind | Why |
|---|---|---|
| `ATTR_ACCESSIBLE` | breaking | Extracted to `protected_attributes` gem; raises `NoMethodError` without the bridge gem |
| `ATTR_PROTECTED` | breaking | Same as above |
| `REQUIRE_STRONG_PARAMS` | breaking | `require 'strong_parameters'` raises `LoadError` — gem now in core |
| `SCOPE_WITHOUT_LAMBDA` | breaking | `scope :foo, where(...)` raises in 4.0; callable required |
| `DEFAULT_SCOPE_WITHOUT_BLOCK` | breaking | Same — block form required |
| `ASSOCIATION_CONDITIONS` | breaking | In-file explanation says "removed"; bridge `activerecord-deprecated_finders` provides warning if installed |
| `ASSOCIATION_ORDER` | breaking | Same |
| `ASSOCIATION_EXTEND` | breaking | Same |
| `ASSOCIATION_UNIQ` | breaking | Same |
| `FINDER_SQL` | deprecation | Explanation explicitly says "deprecated" |
| `FIND_ALL_BY` | deprecation | Explanation says "deprecated in Rails 4.0" |
| `FIND_OR_CREATE_BY` | deprecation | Same |
| `FIND_OR_INITIALIZE_BY` | deprecation | Same |
| `ASSOCIATION_READONLY` | breaking | Explanation: "Rails 4.0 removed :readonly" |
| `RUBY_VERSION` | breaking | 4.0 won't boot on Ruby 1.8 |
| `MATCH_WITHOUT_VIA` | breaking | Raises if `via:` missing |

### MEDIUM priority

| Variable | Kind | Why |
|---|---|---|
| `RESCUE_ACTION` | breaking | "Removed... with no deprecation warning" — `NoMethodError` |
[...]

PR #97, opened 2026-05-05
Document target-hop and dependencies: rules in kind: rubric (closes #76)

### LINKED ISSUE #76: Document 'judge kind at target hop' rule in CLAUDE.md kind: rubric
## Problem

The `## Assigning kind:` rubric in `CLAUDE.md` (added in #54 / #70) defines the four `kind:` values but does not state the meta-rule that governs them: **`kind` is judged at the target hop, not on historical lineage**. The same API can be `deprecation` in one version file and `breaking` in the next.

Surfaced during code review of #73 (rails-40 classification): `SCOPE_WITHOUT_LAMBDA` was deprecated in Rails 3.1 and raises in 4.0. Reviewer asked why it is `breaking` in `rails-40-patterns.yml` rather than `deprecation`. The answer (judge at target hop) is correct but is not written down anywhere — it lives only in the deprecation bullet's "works at this hop" phrase.

## Proposed change

Add a meta-rule paragraph to the `## Assigning kind:` section in `CLAUDE.md`, above the four-value list:

> **Judge `kind` at the target hop, not historical timeline.** A removal that was deprecated in earlier versions is `breaking` in the file for the version where it actually raises. The same API can be `deprecation` in `rails-31-patterns.yml` and `breaking` in `rails-40-patterns.yml`. Apply the same rule to all four kinds: `kind` reflects what the change *is at this hop*, not what it *was* at an earlier hop or *will become* at a later one.

Optionally clarify each of the four values with a "judged at target hop" framing so the rule is reinforced inline:

- `breaking` — Raises / removed / won't boot **at the target version**.
- `deprecation` — Emits a warning **at the target version**; removal scheduled later.
- `migration` — Works today **at the target version**, no warning, recommended migration target.
- `optional` — Opt-in **at the target version**.

### Also document the `dependencies:` ↔ `breaking` relationship

Surfaced during code review of #74 (rails-50). Reviewer asked: does the `migration` kind drive the `dependencies:` section? Answer is no, but it is not written down. Add a short note to `CLAUDE.md` clarifying:

- The top-level `dependencies:` section in each `rails-*-patterns.yml` file lists **bridge / compatibility gems that rescue `breaking` patterns** until the user migrates. Examples: `protected_attributes` rescues `attr_accessible`, `rails-controller-testing` rescues `assigns` / `assert_template`, `activerecord-deprecated_finders` rescues removed dynamic finders.
- `dependencies:` is only relevant to `breaking` kind. The other three kinds resolve in code:
  - `deprecation` → fix before next hop, see per-pattern `fix:` field.
  - `migration` → rewrite call at user's pace, see per-pattern `fix:` field.
  - `optional` → ignore unless user wants the new feature.
- A `breaking` pattern that has a corresponding bridge gem in `dependencies:` is a *softenable* break: the user can install the gem to keep shipping while they migrate. A `breaking` pattern with no bridge entry must be fixed before the version bump.

This relationship is implicit today and was inferred during review rather than read from a doc.

## Why

- Removes ambiguity for future classification PRs (#56 → #66).
- Reviewers can resolve `breaking` vs `deprecation` borderline calls by a single rule rather than re-deriving it case by case.
- Aligns the rubric with how the patterns are already organized: one file per hop, so each pattern is implicitly a statement *about that hop*.

## Out of scope

- Reclassifying any existing `kind:` value. This is a doc-only PR.
- Changes to the validator, output, or top-level key rename.

## References

- Origin: #73 review thread (target-hop rule); #74 review thread (`dependencies:` ↔ `breaking` clarification).
- Rubric: `CLAUDE.md` → "Assigning `kind:`" (added in #54 / #70).
- Parent: #53.

JuanVqz:
## Summary

Closes #76. Documents two rules in `CLAUDE.md`'s `## Assigning kind:` section that have been implicit in the rollout but never written down:

1. **Judge `kind` at the target hop**, not the API's historical timeline.
2. **`dependencies:` ↔ `kind` relationship**: when the `dependencies:` block applies and to which kinds.

Doc-only PR. No code changes. No pattern reclassifications.

### Target-hop rule (added inside `## Assigning kind:`)

A new paragraph above the four-value list states that each `rails-XY-patterns.yml` file is a statement *about that hop* — what changes when the user upgrades INTO that version. The same API can be `deprecation` in `rails-31-patterns.yml` and `breaking` in `rails-40-patterns.yml`. Concrete example called out: `SCOPE_WITHOUT_LAMBDA` (deprecated in Rails 3.1, raises in 4.0 → classified `breaking` in `rails-40-patterns.yml`).

This was the recurring source of "but it was deprecated earlier!" arguments during the classification reviews. The rule is consistent with how the patterns are already organized — one file per hop — but it wasn't written down.

### New section: `## How the dependencies: section relates to kind`

Added immediately after the four-value rubric. Documents that the `dependencies:` block serves **two** distinct purposes, not one:

1. **Bridge / compatibility gems for `breaking` patterns** — e.g., `protected_attributes` rescues `attr_accessible`, `responders` rescues `respond_with`, `rails-controller-testing` rescues `assigns` / `assert_template`. A `breaking` pattern with a corresponding bridge entry is a *softenable* break.
2. **New gems Rails introduces or recommends at this version** — e.g., `bootsnap` (5.2), `propshaft` (8.0), `solid_cache` / `solid_queue` / `solid_cable` (8.0), `kamal` (8.0), `bundler-audit` (8.1). These pair with `optional` patterns or with no pattern at all.

The original issue #76 framed `dependencies:` as "only relevant to `breaking`". That's too strong — the actual data across the 12 pattern files shows new-default gems (typically `optional`) also live in `dependencies:`. The PR uses the more accurate two-purpose framing.

The `check: true` / `check: false` flag is also explained: roughly "you should add this" vs "you may add this".

## Files changed (2 files, +23 lines)

- `CLAUDE.md` — target-hop paragraph + new `## How the dependencies: section relates to kind` section
- `rails-upgrade/CHANGELOG.md` — Unreleased entry

## Test plan

- [x] `bin/validate-patterns` (all 12 files) still passes — no schema changes, doc-only
- [x] No code changes
- [x] No pattern reclassifications

## Linked issues

- Closes #76
- Refs #53 (parent rollout — this is a doc-cleanup follow-up surfaced during PR #73 and PR #74 reviews)
- Refs #94 (sibling SKILL.md threading PR — independent, can land in either order)

## Out of scope

- Reclassifying any `kind:` value (the rule existed implicitly; this PR just writes it down)
- Changes to the validator or any pattern YAML
- SKILL.md updates (covered by #94)


etagwerker (approved):
@JuanVqz Looks good, thanks!

PR #98, opened 2026-05-05
Demote Sprockets/asset patterns to low priority

JuanVqz:
## Summary

Demote three asset-pipeline patterns in `rails-80-patterns.yml` from `high_priority` / `kind: migration` to `low_priority` / `kind: optional`:

- `Sprockets usage`
- `Asset pipeline configuration`
- `javascript_include_tag (missing importmap)`

Also rewrites the `explanation` / `fix` text on each so the optional-migration framing is explicit.

## Why

Sprockets stays a first-class citizen at Rails 8.0. The `sprockets-rails` gem ships and works on 8.0 unchanged. Propshaft is only the *default for newly generated apps*, and migrating an existing Sprockets app to Propshaft is a separate, opt-in modernization, not a fix-before-bump.

The previous shape of these three patterns marked them `kind: migration`, which `workflows/direct-detection-workflow.md` puts in the **fix-before-bump** bucket. On a real 7.2 → 8.0 run against a Sprockets app, that produced a wall of "fix-before-bump" findings (every `config.assets.*` line, every `bootstrap-sprockets` import, every `javascript_include_tag` in a layout) that all map to the same answer: **do nothing**.

The dependency entries at the bottom of `rails-80-patterns.yml` already encode the right posture — `propshaft` and the Solid gems are all `check: false`. This change brings the breaking-change patterns into alignment with that posture.

## Test plan

- [x] `bin/validate-patterns rails-upgrade/detection-scripts/patterns/rails-80-patterns.yml` returns OK.
- [x] Verified on a real 7.2 → 8.0 upgrade: a Sprockets-based app boots and tests pass on Rails 8.0 with no Propshaft migration.

### DIFF
--- a/rails-upgrade/detection-scripts/patterns/rails-80-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-80-patterns.yml
@@ -24,39 +24,6 @@ upgrade_findings:
       fix: "Convert to the explicit form: `enum :status, { active: 0, archived: 1 }, prefix: true`. Mechanical migration; no runtime behavior change once converted"
       variable_name: "ENUM_KWARGS"
 
-    - name: "Sprockets usage"
-      kind: "migration"
-      pattern: "sprockets|Sprockets"
-      exclude: "propshaft"
-      search_paths:
-        - "Gemfile"
-        - "config/"
-        - "app/assets/"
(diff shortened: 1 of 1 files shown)


etagwerker (approved):
@JuanVqz Good idea 👍🏻

PR #99, opened 2026-05-05
Catch gem-level runtime incompat via boot smoke test + pattern prereqs

JuanVqz:
## Summary

Adds two complementary mechanisms to the upgrade workflow for catching gem-level breakage that the current detection and compatibility steps cannot see:

1. **New Step 4.6: Boot Smoke Test on `Gemfile.next`** — between gem-compat (4.5) and report generation (5). Runs a Rails-loading command against the dual-boot file to surface gems that resolve cleanly but blow up at boot.
2. **Optional `prereqs:` field on detection patterns** — declares "the suggested `fix:` only works if gem X is at version Y or above." Lets the report show the cascade up front instead of letting users hit it mid-implementation.

## Motivation

A real 7.1 → 8.0 upgrade ran on a customer app this week. Two fix-before-bump items only surfaced *after* the version flip, because Step 4 (codebase grep) and Step 4.5 (resolver compat check) cannot see them:

| Hop | Gem | Failure | What the existing checks see |
|---|---|---|---|
| 7.2 | `database_cleaner-active_record 2.1.x` | `NoMethodError: undefined method 'schema_migration' for #<...PostgreSQLAdapter>` at first cleaner call | Resolver: compatible (no upper bound on activerecord). Patterns: no app-code reference. |
| 8.0 | `jbuilder 2.11.x` | `LoadError: cannot load such file -- active_support/proxy_object` at `Bundler.require` | Resolver: compatible (no upper bound on activesupport). Patterns: no app-code reference. |

Both gems ship in default Rails-generated apps. The signal that catches them is a Rails boot. Neither static grep nor a constraint resolver will.

Separately, on the same upgrade, the existing `fixture_path` deprecation pattern's `fix:` (rename to `fixture_paths`) silently requires `rspec-rails >= 6.1.0`. On 6.0.x the new setter raises `NoMethodError`. The pattern had no way to express that prereq, so users hit the cascade mid-implementation.

### Step 4.6 + `boot-smoke-test-workflow.md`

- New step in `SKILL.md` between 4.5 and 5. Runs `BUNDLE_GEMFILE=Gemfile.next bundle exec rails runner ...` (or `rspec --dry-run` / `rspec`), diagnoses any `LoadError` / `NoMethodError`, identifies the offending gem, bumps its floor.
- New workflow file at `rails-upgrade/workflows/boot-smoke-test-workflow.md` documenting:
  - Boot trigger options (cheapest → heaviest).
  - Diagnosis recipes (LoadError / NoMethodError patterns).
  - The `find $(bundle show --paths) -name "*.rb" | xargs grep -l ...` recipe for identifying which gem references the missing constant or file.
  - Output format for the report.
  - Two real failure cases (`database_cleaner`, `jbuilder`) as illustrative examples.

The smoke test does **not** replace the post-bump test suite run in Step 6. This is a *boot* check, not a feature check.

### `prereqs:` field on patterns

- Documented in `rails-upgrade/workflows/direct-detection-workflow.md` (pattern format section), with a worked example using the FIXTURE_PATH pattern.
- Compilation rule: when a finding's pattern has `prereqs:`, check `Gemfile.lock`; if a prereq's `min_version` is not met (and `when:` matches), add a fix-before-bump entry for the prereq gem bump **before** the original finding in the report.
- Schema:
  ```yaml
  prereqs:
    - gem: "rspec-rails"
      min_version: "6.1.0"
      reason: "config.fixture_paths= is forwarded from RSpec::Core::Configuration only on rspec-rails 6.1+"
      when: "rspec-rails is present in the bundle"
  ```
- Added to the existing FIXTURE_PATH pattern in `rails-72-patterns.yml` as the first concrete usage.
- `prereqs:` is **optional**. Most patterns describe Rails-only API changes and need none.

## Test plan

- [x] `bin/validate-patterns rails-upgrade/detection-scripts/patterns/rails-72-patterns.yml` returns `OK` with the new `prereqs:` field.
- [x] `SKILL.md` and `direct-detection-workflow.md` render correctly.

## Out of scope

- Adding `prereqs:` to other patterns. Doing one concrete example here; future patterns can add it as needed.
- Automated `Gemfile.lock` parsing for `prereqs:` resolution (the workflow describes the lookup, but doesn't automate it). The skill's existing pattern is "Claude reads the lockfile in the relevant step," which extends naturally.

### DIFF
--- a/rails-upgrade/SKILL.md
+++ b/rails-upgrade/SKILL.md
@@ -170,6 +170,7 @@ If user requests a multi-hop upgrade (e.g., 5.2 → 8.1):
 - `workflows/direct-detection-workflow.md` - How to run breaking change detection directly
 - `workflows/upgrade-report-workflow.md` - How to generate upgrade reports
 - `workflows/gem-compatibility-workflow.md` - **Load in Step 4.5** - Per-lockfile gem compatibility check against the target Rails version. Documents both the primary (`next_rails` `bundle_report compatibility`) and the secondary (railsbump.org API) and the rules for when to escalate.
+- `workflows/boot-smoke-test-workflow.md` - **Load in Step 4.6** - Run a Rails-loading command against `Gemfile.next` to catch gem-level runtime incompat that the resolver can't see (gems calling removed Rails internals or `require`-ing removed files).
 - `workflows/ci-sync-workflow.md` - **MANDATORY before opening the upgrade PR** - How to verify CI config matches the upgraded Gemfile
 - `workflows/app-update-preview-workflow.md` - How to generate app:update previews
 - **`upgrade-cleanup` companion plugin** - User-triggered. Removes dual-boot scaffolding and drops `NextRails.next?` / `NextRails.current?` branches. Deprecation triage stays with this skill for the next hop.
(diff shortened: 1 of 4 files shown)


etagwerker (approved):
@JuanVqz I like this, thanks!

PR #101, opened 2026-05-05
Grep .rspec for deprecation behavior in Step 2

### LINKED ISSUE #100: Silence-check should grep .rspec (and similar autoloaded files) for deprecation behavior overrides
## Context

While exercising the skill on a legacy Rails app, I found a gap in the silence-check guidance for deprecation triage. The skill's current guidance (in workflows that call out where to look for deprecation-silencing settings, e.g. `test-suite-verification-workflow.md` and `references/deprecation-warnings.md`) implicitly directs users to inspect `config/environments/*.rb` and `config/application.rb` for things like `config.active_support.deprecation = :silence` or `:raise`. It does not call out **`.rspec` and other autoloaded config files** (e.g. `Rakefile`, `spec/.rspec`, `spec/spec_helper.rb`, `spec/rails_helper.rb`, Bundler-loaded paths) where deprecation behavior can be installed via `-r` autoloads or test-helper hooks.

> Note: the original report referenced `workflows/deprecation-triage-workflow.md` Step 2. There is no file by that exact name in this repo. The closest matches are `rails-upgrade/workflows/test-suite-verification-workflow.md` (Step 2 = "Run the Test Suite") and `rails-upgrade/references/deprecation-warnings.md` ("Finding Deprecation Warnings"). The fix should land wherever the maintainer feels the silence-check guidance is most discoverable, possibly both.

## What I almost missed

The target app auto-requires `raise_errors_for_deprecations` via its `.rspec` file:

```
# .rspec
-r raise_errors_for_deprecations
```

That single line installs an RSpec hook that raises on every `ActiveSupport::Deprecation.warn`. Without checking `.rspec`, the skill could miss a quietly-installed deprecation-related behavior change (test failures attributed to "the upgrade" when in fact `raise_errors_for_deprecations!` was already turning every warning into a failure pre-upgrade).

## Recommendation

Add `.rspec` and similar autoloaded config files to the grep list for the silence/behavior check. Specifically, sweep:

- `.rspec`
- `spec/.rspec`
- `spec/spec_helper.rb`
- `spec/rails_helper.rb`
- `Rakefile`
- Any Bundler-loaded paths where `RSpec.configure { |c| c.raise_errors_for_deprecations! }` (or equivalent `ActiveSupport::Deprecation.behavior =` overrides) might live (e.g. gems required from the test group, support files under `spec/support/**`).

Suggested grep:

```bash
grep -rnE "raise_errors_for_deprecations|ActiveSupport::Deprecation\.(behavior|disallowed_behavior|silenced)\s*=" \
  .rspec spec/.rspec spec/spec_helper.rb spec/rails_helper.rb Rakefile \
  config/ spec/support/ 2>/dev/null
```

## Why this matters

The silence check is supposed to establish "what is the project's deprecation behavior right now?" before we change it. Missing `.rspec` means we miss an entire category of pre-installed behavior, and we can mis-diagnose post-upgrade test failures.

JuanVqz:
Closes #100
Related https://github.com/ombulabs/claude-code_dual-boot-skill/pull/20

## Summary

- Add a "sweep for deprecation-behavior overrides" substep (2a) to `rails-upgrade/workflows/test-suite-verification-workflow.md` so the silence/raise/disallow check covers `.rspec`, `spec/.rspec`, `spec/spec_helper.rb`, `spec/rails_helper.rb`, `Rakefile`, and `spec/support/`, not just `config/`.
- Mirror the same guidance in `rails-upgrade/references/deprecation-warnings.md` under "Finding Deprecation Warnings", with the canonical `-r raise_errors_for_deprecations` example.

## Note for the maintainer

The originating report referenced `workflows/deprecation-triage-workflow.md` Step 2, but that exact file does not exist in this repo. I landed the change in the two closest spots: `test-suite-verification-workflow.md` (Step 2 = "Run the Test Suite") and `references/deprecation-warnings.md` ("Finding Deprecation Warnings"). Happy to relocate or split if there's a better target.

## Test plan

- [x] Docs-only change. Verified while exercising the skill on a legacy Rails app where `.rspec` autoloaded `raise_errors_for_deprecations` and the original silence-check guidance would have missed it.
- [x] `grep` recipe in the change exercised against that codebase and surfaces the `.rspec` autoload and the `RSpec.configure` hook.

### DIFF
--- a/rails-upgrade/references/deprecation-warnings.md
+++ b/rails-upgrade/references/deprecation-warnings.md
@@ -16,6 +16,27 @@ Deprecation warnings notify you that a specific feature (method, class, or API) 
 
 ## Finding Deprecation Warnings
 
+### Check Existing Deprecation Behavior First
+
+Before reading test logs, sweep for places that already override `ActiveSupport::Deprecation` behavior. Otherwise "no deprecation warnings in the log" can mean "they're being silenced", and "every test fails" can mean "they've been raising the whole time, unrelated to the upgrade." `config/application.rb` and `config/environments/*.rb` are not the only places this is configured: RSpec autoloads `-r` requires from `.rspec`, and test helpers can install hooks like `RSpec.configure { |c| c.raise_errors_for_deprecations! }`.
+
+```bash
+grep -rnE "raise_errors_for_deprecations|ActiveSupport::Deprecation\.(behavior|disallowed_behavior|silenced)\s*=|ActiveSupport::Deprecation\.silence\b" \
+  .rspec spec/.rspec spec/spec_helper.rb spec/rails_helper.rb Rakefile \
+  config/ spec/support/ lib/tasks/ 2>/dev/null
+```
+
+The trailing `silence\b` alternation catches the block form `ActiveSupport::Deprecation.silence do ... end` (or `{ ... }`), which is the recommended way to scope a silence to a single noisy call site. The `\b` word boundary keeps it from matching `silenced` (which is already covered by the assignment branch).
+
+A common gotcha is a single line in `.rspec`:
+
+```
+# .rspec
+-r raise_errors_for_deprecations
+```
+
+That installs an RSpec hook that raises on every `ActiveSupport::Deprecation.warn`. Easy to miss because nothing about it lives under `config/`.
+
 ### In Test Suite Logs
 
 If you have good test coverage, run your test suite and check the logs:
--- a/rails-upgrade/references/deprecation-warnings.md
+++ b/rails-upgrade/references/deprecation-warnings.md
@@ -18,15 +18,24 @@ Deprecation warnings notify you that a specific feature (method, class, or API) 
 
 ### Check Existing Deprecation Behavior First
 
-Before reading test logs, sweep for places that already override `ActiveSupport::Deprecation` behavior. Otherwise "no deprecation warnings in the log" can mean "they're being silenced", and "every test fails" can mean "they've been raising the whole time, unrelated to the upgrade." `config/application.rb` and `config/environments/*.rb` are not the only places this is configured: RSpec autoloads `-r` requires from `.rspec`, and test helpers can install hooks like `RSpec.configure { |c| c.raise_errors_for_deprecations! }`.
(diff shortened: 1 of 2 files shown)


etagwerker:
Thanks for catching the `.rspec` autoload trap, that one is easy to miss.

I have a few patterns the recipe does not currently surface, and I would love your take on whether any of them feel worth adding or out of scope.

1. The env DSL form, `config.active_support.deprecation = :silence` in `config/environments/*.rb`, has been the canonical way to configure deprecation behavior since Rails 3, and I suspect it is the form most apps actually use. Would you be open to extending the regex with `config\.active_support\.(deprecation|disallowed_deprecation|disallowed_deprecation_warnings|report_deprecations)\s*=`?

2. The Rails 7.1+ deprecators registry: `Rails.application.deprecators.behavior =`, `.silence do`, plus per-framework deprecators like `ActiveRecord.deprecator.silence do`. Since the skill goes up through 8.1, would it make sense to include those? Something like `(Rails\.application\.deprecators|[A-Z]\w*\.deprecator)\.(behavior|disallowed_behavior|silenced)\s*=` with the matching `\.silence\b` form.

3. Minitest-shaped apps. The current paths are RSpec-centric. For apps without `spec/`, would it help to also sweep `test/test_helper.rb`, `test/support/`, and `test/minitest/`? The `2>/dev/null` already absorbs "no such file" noise.

4. A short version-context note. Would a one-liner mapping each pattern to the Rails version where it appeared help readers interpret results? Mostly thinking of someone running this against a Rails 5 app who would otherwise wonder why no `deprecators` calls turn up.

One open question: if a single comprehensive recipe feels like it is collapsing too many concerns into one grep, would a tiered approach read more cleanly to you, a baseline check plus a follow-up for 7.1+ apps? I am leaning toward the single recipe since `grep` is cheap, but happy to defer.

etagwerker:
Went ahead and pushed my own suggestions into the branch so this isn't blocked on you, @JuanVqz — shout if any of it feels off and I'll happily revert or adjust.

Folded all four into the single Step 2 recipe (kept it as one grep rather than tiering it, since grep is cheap and one block is easier to keep in sync across the two files):

1. **env DSL form** — added the `config.active_support.deprecation` / `disallowed_deprecation` / `disallowed_deprecation_warnings` / `report_deprecations` branch. Agreed this is probably the form most apps actually use.
2. **Rails 7.1+ deprecators registry** — folded `Rails.application.deprecators` and per-framework `<Framework>.deprecator` (e.g. `ActiveRecord.deprecator`) into the `behavior` / `disallowed_behavior` / `silenced` / `.silence` branches. Verified it catches all three shapes.
3. **Minitest apps** — added `test/test_helper.rb`, `test/support/`, `test/minitest/` to the path list.
4. **Version context** — added a short "which forms appear depends on Rails version" note so someone running this against a Rails 5–7.0 app sees "no `deprecators` hits" as expected rather than a gap.

I sanity-checked the regex against fixtures for each form, and confirmed a bare `:silenced` symbol does not false-positive. Ready for another look whenever you have a minute.

etagwerker (approved):

PR #105, opened 2026-05-26
Rails 4.0 patterns: validation anchors + accepts_nested_attributes_for

JuanVqz:
## Summary

Three new Rails 4.0 detection patterns. All three break boot when the model class loads, and the current skill does not catch them.

- **`VALIDATES_FORMAT_OF_ANCHORS`** — Rails 4 raises `ArgumentError: The provided regular expression is using multiline anchors (^ or $)...` when a `validates_format_of` regex starts with `^` or ends with `$`. This check is new in 4.0 (Rails 3.2 does not have it), checked against `format.rb` at v3.2.22.5 vs v4.0.0. Fix is `\A` / `\z`, works on both versions, no gate. The regex now matches both anchors (`[\^$]`), so trailing-`$`-only forms like `with: /foo$/` are caught. Only a leading `^` or trailing `$` actually raises, so a `^` inside a character class (`/[^@]+/`) is a false positive — noted in the entry. Block-options and method-call forms are not caught by the main regex; a second grep is given for them.
- **`VALIDATES_FORMAT_OF_ANCHORS_CONSTANT_BOUND`** — catches regex constants built with `Regexp.new(%q{^...$})` that feed validators, which the inline grep misses. The regex no longer stops at the first `)`, so grouped patterns like `Regexp.new("(a|b)$")` are caught. Check each hit by hand — not every `Regexp.new` constant feeds a validator, and only a leading `^` or trailing `$` raises.
- **`ACCEPTS_NESTED_NO_REFLECTION`** — `accepts_nested_attributes_for :foo` raises `ArgumentError: No association found for name 'foo'` at class-load when `:foo` has no association. This raise is not new in Rails 4.0 (it fires on any version when the association is missing), so a missing association is not a 3.2 → 4.0 regression. It shows up as a new failure on the next side only when the association comes from a gem's base class through inheritance, and a gem bump in `Gemfile.next` removes it. Example: paper_trail 2.x has a top-level `Version` (with `belongs_to :item`); 3.x renames it to `PaperTrail::Version` and drops the top-level `Version`, so models written as `< Version` lose the inherited association on the next side only. A hit is only a problem when the association is inherited from a gem's base class and that gem is being bumped in `Gemfile.next`. Two fixes given: re-point the base class on the next side in one place (`Version = PaperTrail::Version if NextRails.next?`), or add the association per model (`belongs_to ... if NextRails.next?`, gated so the current side does not get a duplicate reader/writer). A hand-written `def foo` override still wins either way.

All three classified `high_priority` / `kind: breaking`.

## Test plan

- [ ] Reviewer spot-checks the three regexes against representative 3.x code (literal `validates_format_of`, `Regexp.new` constants, `accepts_nested_attributes_for` macros).


etagwerker (approved):
LGTM. I made a small adjustment to support => syntax.

PR #107, opened 2026-05-26
Rails 4.0 patterns: bridge gem swaps (paper_trail, paranoia) + actionpack-page_caching dep

JuanVqz:
## Summary

Two Rails 4.0 bridge-gem swap patterns plus a missing dependencies registry entry.

- **`PAPER_TRAIL_STRING_VERSIONS`** — new entry. `has_paper_trail` internally calls `has_many` with the name from `versions:` / `version:`. On Rails 4, when the option value is a string, paper_trail's own code raises `ArgumentError: association names must be a Symbol`. Boot-fatal. Fix: change `versions: "string"` → `versions: :symbol` (paper_trail accepts symbols on 3.2 too, so the rewrite ships pre-bump).
- **`VALIDATES_AS_PARANOID`** — new entry. `validates_as_paranoid` is a class-method DSL provided by `rails3_acts_as_paranoid` (Rails-3-only soft-delete gem). Typical 3.2 → 4.0 hop gates the gem `unless next?` in the Gemfile because it's unmaintained on Rails 4+. Call sites then raise `NoMethodError` at model class-load on the next side. The `paranoia` replacement gem does NOT provide `validates_as_paranoid`; its default_scope auto-excludes soft-deleted records from `Model.where(...)` uniqueness validations. Fix: `validates_as_paranoid unless NextRails.next?`.
- **`actionpack-page_caching` added to dependencies registry**, and the existing `actionpack-action_caching` message scoped to `caches_action` only. At Rails 4.0 page caching and action caching were extracted into two separate gems with distinct release lines — the existing OSS entry conflated them.

Both new patterns classified `high_priority` / `kind: breaking`.

## Test plan

- [ ] Reviewer confirms PAPER_TRAIL regex matches both `versions:` and `version:` (singular/plural).
- [ ] Reviewer sanity-checks the actionpack-page_caching URL and message.


rails-upgrade/detection-scripts/patterns/rails-40-patterns.yml
```diff
@@ -368,11 +388,23 @@ dependencies:
...
+    message: "Required if using caches_page — extracted from Rails core at 4.0 (separate gem from actionpack-action_caching)"
+    url: "https://github.com/rails/actionpack-page_caching"
+
   activeresource:
     check: false
     gem_name: "activeresource"
     message: "Required if using ActiveResource"
     url: "https://github.com/rails/activeresource"
+
+  paper_trail:
```
JuanVqz, line 539: I'm unsure of adding this here because, yes, this is the right movement, but I'm not sure how to use it in the skill.

etagwerker (approved):

PR #109, opened 2026-05-26
Rails 4.0 patterns: internal API + middleware edges (CollectionProxy, HWIA, Rack::Lock)

JuanVqz:
## Summary

Three new Rails 4.0 internal-API + middleware patterns. All `medium_priority`, bounded but real classes of apps. Every claim verified against rails/rails source at tags v3.2.22 and v4.0.0.

- **`COLLECTION_PROXY_ENUMERABLE_GUARD`** (`breaking`) — Rails 4.0 rewrote `CollectionProxy` to inherit from `ActiveRecord::Relation`, and 4.0 `Relation` does NOT `include Enumerable` (it delegates enumerable methods to `to_a`). On 3.2 `CollectionProxy` was a blank-slate proxy that forwarded `is_a?` to its underlying Array via `method_missing`, so `proxy.is_a?(Enumerable)` returned true. Result: code using `value.is_a?(Enumerable)` to tell a single record from a collection classifies a has-many proxy as a collection on 3.2 but as a single record on 4.0. Silent semantic shift, no error. Fix broadens the guard to `value.is_a?(Enumerable) || value.is_a?(::ActiveRecord::Relation)` (leading `::` required to avoid `NameError` under a shadowing scope). Verified against source; the explicit `Relation` check keeps the fix correct even though later Rails re-added `Enumerable` to `Relation`.

- **`HWIA_CONVERT_VALUE_OVERRIDE`** (`breaking`) — Rails 4.0.0 changed `HashWithIndifferentAccess#convert_value` to `convert_value(value, options = {})`, and internal callers pass a second arg (`for: :assignment` / `for: :to_hash`). A 1-arg override `def convert_value(value)` then raises `ArgumentError: wrong number of arguments (given 2, expected 1)`. Confirmed the second parameter landed in 4.0.0 (not 4.1/4.2). Fix adds `_options = {}` as a second parameter, backwards-compatible.

- **`RACK_LOCK_MIDDLEWARE`** (`breaking`) — Rails 4.0 did not remove Rack::Lock from the default stack; it changed the conditional. `allow_concurrency?` falls back to `config.cache_classes` when `allow_concurrency` is nil (the 4.0 default), so Rack::Lock is skipped wherever `cache_classes` is true (production, and the test env under CI). Any MiddlewareStack call naming it as a target then raises at boot via `assert_index`: `RuntimeError: No such middleware to insert before: Rack::Lock` (verbatim, or `after`). Per stack.rb at v4.0.0 the raising methods are `insert` (alias `insert_before`), `insert_after`, and `swap`; `use` / `unshift` / `delete` do not raise. The pattern matches all four raising forms (space and paren styles) and excludes the integer-index `insert(0, ...)`. Fix is `config.allow_concurrency = false` in the affected env (usually test.rb): it reinstates Rack::Lock so the original `insert_before` / `insert_after` line works unchanged on both versions, no `NextRails.next?` branch, same slot, and the one setting fixes both forms identically. On 3.2 that value is already the default, so it's a no-op there. Verified with `rake middleware` on both the current and Gemfile.next bundles that the middleware lands in the same slot.

## Test plan

- [x] `RACK_LOCK_MIDDLEWARE` regex matches the four raising forms (`insert` / `insert_before` / `insert_after` / `swap` before `Rack::Lock`, space and paren styles) and excludes `insert(0, ...)`, `delete`, `unshift`, `use`.
- [x] All three claims verified against rails/rails v3.2.22 + v4.0.0 source (CollectionProxy/Relation Enumerable, convert_value 4.0.0 signature, Rack::Lock conditional insertion + error wording).


etagwerker (approved):
👍🏻

PR #110, opened 2026-05-27
Add fixture-test harness for detection patterns (seed: rails-40)

JuanVqz:
@etagwerker just checking if we are interested in having "tests" to validate our patterns.

## Why

`bin/validate-patterns` confirms each detection pattern's regex *compiles* and that the file matches the schema. Nothing confirmed a regex actually *matches what its explanation claims* or *skips what it shouldn't*. "What does this pattern cover" lived only in prose, so coverage gaps stayed invisible until a real upgrade missed a breaking change.

PR #105 showed this concretely: two anchor patterns had false negatives caught only by hand during review. A fixture test would have caught them immediately.

## What this adds

- **`bin/test-patterns`** — pure-stdlib runner, same shape as `bin/validate-patterns` (including `--self-test`). For each pattern file it loads a sibling `*.expectations.yml` and asserts every `match` line is flagged and every `no_match` line is not. A line is "flagged" when the pattern regex matches AND the `exclude` regex (if any) does not, approximating the skill's per-line grep + grep -v.
- **`rails-40-patterns.expectations.yml`** — the expectations format. The `match` list doubles as a living scope doc: find an uncovered shape, add it to `match`, the test goes red, then widen the regex.
- **CI** — `test-patterns` and its self-test run alongside `validate-patterns`.
- **`validate-patterns`** now skips `*.expectations.yml` so it does not schema-check them.

## Scope: start with one

This intentionally seeds **one** pattern (`SCOPE_WITHOUT_LAMBDA`) as the worked template, so the approach is easy to review. If this lands, we replicate the same `match` / `no_match` shape across the rest of the rails-40 patterns and the other version files.

## Verification

- `bin/test-patterns --self-test` → 5/5
- `bin/test-patterns` → rails-40 OK (4 assertions, 1/32 patterns covered), other files SKIP (no expectations yet)
- `bin/validate-patterns` + `--self-test` → green
- Malformed expectations YAML reports a clean `FAIL ... YAML parse error` (exit 1), not a backtrace

### DIFF
--- a/.github/workflows/validate-patterns.yml
+++ b/.github/workflows/validate-patterns.yml
@@ -6,11 +6,13 @@ on:
     paths:
       - "rails-upgrade/detection-scripts/patterns/*.yml"
       - "bin/validate-patterns"
+      - "bin/test-patterns"
       - ".github/workflows/validate-patterns.yml"
   pull_request:
     paths:
       - "rails-upgrade/detection-scripts/patterns/*.yml"
       - "bin/validate-patterns"
+      - "bin/test-patterns"
       - ".github/workflows/validate-patterns.yml"
 
 permissions:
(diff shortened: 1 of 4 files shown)


etagwerker (approved):
@JuanVqz I like it, thanks!

PR #115, opened 2026-06-05
Add Rails 8.0 undocumented breaking change: belongs_to association query_constraints removed in favor of foreign_key

fullmetalengineer:
## What

Adds the `query_constraints:` association option removal to the **7.2 → 8.0** upgrade guidance, in two places:
  
- `version-guides/upgrade-7.2-to-8.0.md` — new HIGH PRIORITY breaking-change section with detection pattern, the exact `ConfigurationError`, and an OLD/NEW fix.
- `detection-scripts/patterns/rails-80-patterns.yml` — a `high_priority` / `kind: breaking` entry (`QUERY_CONSTRAINTS_OPTION`) so the change is auto-detected.

## Why

Rails 8.0 **removes** the `query_constraints:` option on associations (`belongs_to`/`has_many`/etc.). It was deprecated in 7.2 and now **raises `ActiveRecord::ConfigurationError` at class load**, so any model using it prevents the app from booting:

```
ActiveRecord::ConfigurationError:
  Setting `query_constraints:` option on `ActivityParticipant.belongs_to :activity`
  is not allowed. To get the same behavior, use the `foreign_key` option instead.
```

The fix is behavior-preserving and version-agnostic — pass the composite key as an `Array` to `foreign_key:`:

```ruby
# BEFORE (7.2 deprecation → 8.0 raises)
belongs_to :activity, query_constraints: [:activity_id, :school_id]

# AFTER (works on Rails 7.2 AND 8.0)
belongs_to :activity, foreign_key: [:activity_id, :school_id]
```

ActiveRecord maps an `Array` `foreign_key` back to `query_constraints` internally on **both** 7.2 and 8.0, so no `NextRails.next?` dual-boot branch is needed, and it also silences the 7.2 deprecation warning.        
  
## Why it was missing

This breaking change is **not documented in the official Rails Upgrade Guide or the 7.2/8.0 release notes** — only in the `activerecord` CHANGELOG (7.2) and the source. It surfaced during a real 7.2 → 8.0 upgrade via a **boot smoke
  test** (`bin/rails runner`), which raised the `ConfigurationError` on class load; an upgrade docs only review would have missed it.

## Detection-pattern note

The pattern is intentionally the broad `query_constraints:` rather than an association-anchored regex. For a `kind: breaking` boot failure, recall matters more than precision: the broad form also catches multi-line `belongs_to` declarations (where the option is on a continuation line), at the cost of an occasional false positive on an unrelated `query_constraints:` hash key, which is cheap to dismiss in the verification step. It correctly skips the legitimate composite-PK class macro `query_constraints :col1, :col2` (space before the symbol, no adjacent colon).

## Sources

- `activerecord-7.2.3/CHANGELOG.md:620` — *"Association option `query_constraints` is deprecated in favor of `foreign_key`."* (Nikita Vasilevsky)
- `activerecord-7.2.3/CHANGELOG.md:809-810` — `Array` `foreign_key` replaces the composite-key use of `query_constraints`.
- `activerecord-8.0.5/lib/active_record/reflection.rb:525-530` — `raise ConfigurationError` on `query_constraints:`.
- `reflection.rb` (7.2 `:541` and 8.0 `:533`) — `if options[:foreign_key].is_a?(Array): options[:query_constraints] = options.delete(:foreign_key)` — the `Array` → `query_constraints` mapping, identical in both versions.

### DIFF
--- a/rails-upgrade/detection-scripts/patterns/rails-80-patterns.yml
+++ b/rails-upgrade/detection-scripts/patterns/rails-80-patterns.yml
@@ -44,6 +44,16 @@ upgrade_findings:
       fix: "Remove sqlite3_deprecated_warning configuration"
       variable_name: "SQLITE3_WARNING"
 
+    - name: "belongs_to/has_many query_constraints: option"
+      kind: "breaking"
+      pattern: "query_constraints:"
+      exclude: ""
+      search_paths:
+        - "app/models/"
+      explanation: "Rails 8.0 removes the `query_constraints:` association option (deprecated in 7.2; activerecord/CHANGELOG.md 7.2: 'Association option `query_constraints` is deprecated in favor of `foreign_key`.'). At 8.0 it raises `ActiveRecord::ConfigurationError` at class load, so the model — and the app — fails to boot. Not mentioned in the official upgrade guide or release notes. Verify the match is an association option, not an unrelated query_constraints: hash key."
+      fix: "Pass the composite key as an Array to `foreign_key:` instead: `belongs_to :activity, foreign_key: [:activity_id, :school_id]`. AR maps an Array `foreign_key` to query_constraints internally on both 7.2 and 8.0, so the change is behavior-preserving and needs no dual-boot branch."
+      variable_name: "QUERY_CONSTRAINTS_OPTION"
+
   medium_priority:
     - name: "Redis cache store"
       kind: "optional"
--- a/rails-upgrade/version-guides/upgrade-7.2-to-8.0.md
+++ b/rails-upgrade/version-guides/upgrade-7.2-to-8.0.md
@@ -163,9 +163,42 @@ rbenv local 3.3.0
 
 ---
 
+#### 6. `query_constraints:` association option removed (composite foreign keys)
+
+**What Changed:**
+Rails 8.0 **removes** the `query_constraints:` option on associations (`belongs_to`/`has_many`/etc.). It was deprecated in Rails 7.2 and now raises `ActiveRecord::ConfigurationError` **at class load**, so the app fails to boot.
+
+**Detection Pattern:**
+```ruby
+# app/models/*.rb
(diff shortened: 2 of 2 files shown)


JuanVqz:
Hey @fullmetalengineer, thanks for opening the PR, and sorry for the delay.

I'm going to review it and get back to you

JuanVqz:
Okay, this looks good:

From my investigation this addition adds value, and we want to merge it, a bit of more context:

This was first introduced at Rails 7.1 as part of the composite primary key support.

Let you point an association at a composite FK: `belongs_to :author, query_constraints: [:author_first_name, :author_last_name]`

Why deprecated (7.2) → removed (8.0): redundancy.
  - 7.1 also taught `foreign_key` to accept an Array for the same composite-FK case.
  - Two public options, one job. Rails consolidated on `foreign_key:`.

`query_constraints` stayed as the internal mechanism; the public spelling became `foreign_key: [...]`.
Deprecating the explicit option cost nothing behaviorally: the same engine underneath.
That is also why your PR's fix is behavior-preserving across 7.2 and 8.0.

## Timeline

```
7.1 introduced → 7.2 deprecated (warning) → 8.0 raises `ConfigurationError`
```

## Sources
  - [Rails 7.1 Composite Primary Keys guide](https://guides.rubyonrails.org/v7.1/active_record_composite_primary_keys.html)
  - [Saeloun: Rails 7.1 Composite Primary Key Support](https://blog.saeloun.com/2024/08/06/active-record-composite-primary-key/)
  - [Rails 7.1 Release Notes](https://guides.rubyonrails.org/7_1_release_notes.html)


Thank you, @fullmetalengineer, for the work; keep the contributions coming.

JuanVqz (approved):

fullmetalengineer:
Appreciate the merge here — sorry I didn't reply to you earlier. Been working on some extraordinarily time-sensitive things at work and this took a back seat.

Happy to contribute! Thanks for putting this skill together.

Issue #113, opened 2026-06-02
Projects with no test suite


hmdros:
Can we have the skill "smoke test" for us when there is no test suite?

Some applications has zero or low test coverage and would be interesting the skill to gather most known breakpoints and try to run a background check for that.

etagwerker:
@hmdros Does #116 solve this issue for you?

PR #116, opened 2026-06-08
docs: add no-test-suite smoke fallback

beanscg:
## Summary

Adds a concrete fallback for #113 when a Rails app has no runnable RSpec or Minitest suite.

Changes:

- Adds `workflows/no-test-suite-smoke-workflow.md` with safe read-only baseline checks for Rails boot, test-env boot, routes, migration status, and asset/build commands
- Wires the fallback into Step 1 of `rails-upgrade/SKILL.md`
- Updates the test-suite verification workflow so "no tests found" records a partial-confidence smoke baseline instead of only warning and proceeding

This keeps the existing "real test suite first" rule intact: the fallback only applies when no runnable suite exists, and boot/routes failures still stop the upgrade.

## Validation

- `docker run --rm -v /Users/stephenkall/Desktop/BEAN/.tmp/bean-gh488-rails-skill-validate:/app -w /app ruby:3.3-alpine sh -lc 'ruby bin/validate-patterns && ruby bin/validate-patterns --self-test'` -> all pattern files OK, self-test 8/8
- `git diff --check`
- `rg -n "no-test-suite-smoke|No-Test-Suite|no runnable test suite|baseline confidence" rails-upgrade/SKILL.md rails-upgrade/workflows/test-suite-verification-workflow.md rails-upgrade/workflows/no-test-suite-smoke-workflow.md`

Note: local macOS system Ruby 2.6 cannot run `bin/validate-patterns` because `Psych.safe_load_file` is unavailable, so validation was run with Ruby 3.3 in Docker.

Submitted by Bean Labs as a public free-service contribution; acceptance is still pending maintainer review.

### DIFF
--- a/rails-upgrade/workflows/no-test-suite-smoke-workflow.md
+++ b/rails-upgrade/workflows/no-test-suite-smoke-workflow.md
@@ -1,0 +1,119 @@
+# No-Test-Suite Smoke Workflow
+
+**Purpose:** Provide a concrete baseline check when a Rails app has no RSpec or Minitest suite.
+
+**When to use:** Step 1 of the Rails upgrade workflow, only after test-suite detection finds no runnable `spec/` or `test/` suite.
+
+This workflow is not a replacement for adding tests. It is a minimum boot and routing baseline so the upgrade report can say exactly what was checked before proceeding.
+
+---
+
+## Step 1: Confirm There Is No Runnable Suite
+
+Check both files and Gemfile entries before declaring "no tests":
+
+```bash
+test -d spec || test -d test
+grep -E "rspec-rails|minitest|capybara|factory_bot_rails" Gemfile Gemfile.lock
+```
+
+If there are test directories or gems, return to `test-suite-verification-workflow.md` and run the real suite. Use this fallback only when no runnable suite exists.
+
+---
+
+## Step 2: Run Rails Boot Checks
+
+Start with the cheapest command that loads the Rails environment:
+
+```bash
+bundle exec rails runner "puts Rails.version"
+```
+
+If the app has multiple environments or credentials constraints, also check test-mode boot:
+
+```bash
+RAILS_ENV=test bundle exec rails runner "puts Rails.env"
+```
+
+Record PASS or FAIL and the exact error. A boot failure blocks the upgrade until the user decides whether to fix the baseline first.
+
+---
+
+## Step 3: Check Routes And Migrations
+
+These commands catch common baseline failures without needing a test suite:
+
+```bash
+bundle exec rails routes >/tmp/rails-routes.txt
+bundle exec rails db:migrate:status
+```
+
+If `db:migrate:status` cannot run because the database is unavailable, report it as "not verified" rather than fabricating a pass. Do not create or migrate a database unless the user asked for that setup.
+
+---
+
+## Step 4: Run Asset Or Build Checks When Available
+
+Only run commands that already exist in the app:
+
+```bash
+bundle exec rails assets:precompile
+bin/vite build
+yarn build
+npm run build
+```
+
+Pick the app's actual build path from its files (`package.json`, `vite.config.*`, `app/assets`, `config/webpacker.yml`, or Propshaft/Sprockets config). If no asset build exists, record "not applicable."
+
+---
+
+## Step 5: Optional Manual Smoke URLs
+
+If the app can boot locally, collect the most important URLs from routes:
+
+```bash
+bundle exec rails server
+```
+
+Then manually or with curl check only safe read-only endpoints, for example:
+
+```bash
+curl -I http://localhost:3000/
+curl -I http://localhost:3000/users/sign_in
+```
+
+Do not submit forms, mutate data, or hit admin/customer actions as part of this fallback.
+
+---
+
+## Output Format
+
+```markdown
+## No-Test-Suite Smoke Baseline
+
+**Status:** PASS / FAIL / PARTIAL
+**Reason this fallback ran:** No runnable RSpec or Minitest suite was detected.
+
+| Check | Command | Result |
+|---|---|---|
+| Rails boot | `bundle exec rails runner "puts Rails.version"` | PASS |
+| Test env boot | `RAILS_ENV=test bundle exec rails runner "puts Rails.env"` | PASS |
+| Routes load | `bundle exec rails routes` | PASS |
+| Migration status | `bundle exec rails db:migrate:status` | NOT VERIFIED - database unavailable |
+| Asset/build check | `<actual command>` | PASS / NOT APPLICABLE |
+
+**Upgrade risk:** High. This app has no automated suite. Add focused tests before or during the first upgrade hop, and treat every generated change as requiring manual verification.
+```
+
+---
+
+## Stop Conditions
+
+Stop and report instead of proceeding when:
+
+- Rails cannot boot in the current environment.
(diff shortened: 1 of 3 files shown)


etagwerker (approved):
@beanscg I like this idea! 

what do you think, @JuanVqz?

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
```diff
@@ -0,0 +1,119 @@
...
+Record PASS or FAIL and the exact error. A boot failure blocks the upgrade until the user decides whether to fix the baseline first.
+
+---
+
+## Step 3: Check Routes And Migrations
+
+These commands catch common baseline failures without needing a test suite:
+
+```bash
+bundle exec rails routes >/tmp/rails-routes.txt
```
JuanVqz: 1. Only stdout is redirected. Rails routing errors (e.g., a bad constant in `routes.rb`) print to stderr and are silently discarded. The file ends up empty, but the agent may still record PASS if it only checks whether the command ran.

2. `/tmp` is not portable. Sandboxed CI environments, Docker containers with a read-only `/tmp`, and macOS App Sandbox contexts can all fail this write silently or noisily. The routes check then appears to fail for the wrong reason.

Let's capture both streams, use a project-relative path:

```bash
bundle exec rails routes > .smoke-routes.txt 2>&1 # or a tmp/ folder
# ... record result ...
```

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
```diff
@@ -0,0 +1,119 @@
...
+Pick the app's actual build path from its files (`package.json`, `vite.config.*`, `app/assets`, `config/webpacker.yml`, or Propshaft/Sprockets config). If no asset build exists, record "not applicable."
+
+---
+
+## Step 5: Optional Manual Smoke URLs
+
+If the app can boot locally, collect the most important URLs from routes:
+
+```bash
+bundle exec rails server
```
JuanVqz: I think we should include shutting the server down, or agents might leave it running and cause conflicts at some point.

```bash
bundle exec rails server &
SERVER_PID=$!
sleep 5  # wait for boot
curl -I http://localhost:3000/
kill $SERVER_PID 2>/dev/null
```

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
```diff
@@ -0,0 +1,119 @@
...
+
+---
+
+## Step 1: Confirm There Is No Runnable Suite
+
+Check both files and Gemfile entries before declaring "no tests":
+
+```bash
+test -d spec || test -d test
+grep -E "rspec-rails|minitest|capybara|factory_bot_rails" Gemfile Gemfile.lock
```
JuanVqz: I see two problems with this grepping:

1. `minitest` is included by default when a Rails app is generated, even when the developer never wrote a single test. An app that was created with `rails new` and immediately abandoned its test setup will still match this grep and get sent back to `test-suite-verification-workflow.md`, where no runnable suite is found, and then bounced back here. Infinite loop.

2. Searching `Gemfile.lock` compounds the problem: `minitest` appears as a transitive dependency of `activesupport` in virtually every Rails `Gemfile.lock`, regardless of intent.

The detection needs to check for actual test files, not just gem presence. A test gem with no test files means no runnable suite.

Let's check for the presence of `spec/test` files alongside the directory and gem check:

```bash
# Directory check
test -d spec && find spec -name "*_spec.rb" | grep -q .
test -d test && find test -name "*_test.rb" | grep -q .

# Gemfile only (not Gemfile.lock), explicit test gems only
grep -E "rspec-rails|minitest-rails" Gemfile
```

Treat the app as having no runnable suite unless at least one test file is found. agree?

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
```diff
@@ -0,0 +1,119 @@
...
+```bash
+bundle exec rails routes >/tmp/rails-routes.txt
+bundle exec rails db:migrate:status
+```
+
+If `db:migrate:status` cannot run because the database is unavailable, report it as "not verified" rather than fabricating a pass. Do not create or migrate a database unless the user asked for that setup.
+
+---
+
+## Step 4: Run Asset Or Build Checks When Available
```
JuanVqz, line 61: I think there is an `assets:check` command we might use, but I'm not sure from what rails version it exists.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
```diff
@@ -0,0 +1,119 @@
...
+
+---
+
+## Stop Conditions
+
+Stop and report instead of proceeding when:
+
+- Rails cannot boot in the current environment.
+- Routes cannot load.
+- The app has a test suite but it fails to run.
```
JuanVqz: This workflow is only entered after test-suite detection has already confirmed no runnable suite exists. so, this might be moved into `test-suite-verification-workflow.md`

JuanVqz (commented):
Hey @beanscg Thanks for the contribution! I had a look at the PR and found some issues worth addressing before we ship this.

Ping me on any idea I shared and you disagree with; we can chat about it.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
etagwerker: Good catch, agreed. Switched detection to test-file presence (`find spec -name "*_spec.rb"` / `test -name "*_test.rb"`) as the deciding signal, and limited the gem grep to `Gemfile` only with explicit test gems (`rspec-rails|minitest-rails`). No more `Gemfile.lock` matching, so the transitive-minitest infinite bounce is gone.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
etagwerker: Done. Now `bundle exec rails routes > tmp/rails-routes.txt 2>&1` so stderr routing errors surface instead of leaving a silent empty file, and it writes to the app-local `tmp/` instead of `/tmp` to avoid read-only/sandboxed failures.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
etagwerker: Agreed, added the background-server + `SERVER_PID` + `kill $SERVER_PID` pattern so we never leave an orphaned server holding port 3000.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
etagwerker: Right, that case is unreachable here since we only enter after detection confirms no runnable suite. Removed it from Stop Conditions. The "suite fails to run" case belongs in `test-suite-verification-workflow.md`.

rails-upgrade/workflows/no-test-suite-smoke-workflow.md
etagwerker: I looked into this and there is no reliable cross-version `assets:check` task. The stable ones are `assets:precompile` / `assets:clobber`; some Sprockets versions expose `assets:environment` but nothing like a generic check, and it varies across the Rails versions this skill targets. I left Step 4 on `assets:precompile` rather than add a task that may not exist. Happy to revisit if you know a version where `assets:check` is dependable.

PR #128, opened 2026-08-05
Harden scope/default_scope detection across the 4.x hops

JuanVqz:
## Why

A non-callable scope body is not a style issue on Rails 4.0. It silently returns wrong rows, and the skill treated it as a mechanical rewrite.

Rails 3.2 merged the stored relation into the current scope at call time, so an association's condition survived. Rails 4.0 returns it verbatim on the non-callable branch — no `merge`, no `scoping` — so everything to the left is discarded:

```sql
-- 3.2, and 4.0 with `scope :active, -> { where(deleted: nil) }`
WHERE `issues`.`incident_id` = 123 AND `issues`.`deleted` IS NULL

-- 4.0 with `scope :active, where(deleted: nil)`
WHERE `issues`.`deleted` IS NULL
```

Rails only warns, at class-definition time. This shipped to production in a real 3.2 → 4.0 upgrade: a page called `incident.issues.active` to decide whether a record was editable, the unbounded read killed the worker, and it surfaced as a 503. Dev and CI held too little data for the query to be slow, so every pre-production gate passed.

The old detection regex looked only for `where`, so it also missed `order(...)`, `joins(...)` and `scoped.where(...)`.

## What changed

**Widened detection.** `SCOPE_WITHOUT_LAMBDA` now matches any non-callable body, and `DEFAULT_SCOPE_WITHOUT_BLOCK` any non-block argument (it was `where|order|:order`, missing `default_scope joins(...)` and friends). Both excludes are anchored to the body position and tolerate a wrapping paren.

**New patterns.** `SCOPE_BODY_CONTINUATION` for a body that wraps onto the next line, which a line-oriented pattern cannot classify — flagged for human review rather than dropped. `SCOPE_NAME_COLLISION` for the name guard 4.1 added to `#scope`; `scope :none` is the usual casualty, since Active Record gained `Model.none` in 4.0 and a hand-rolled 3.2-era scope now collides.

**Per-hop coverage.** The same defect has a different symptom at each version, so each hop's pattern file and guide carries its own entry:

| Hop | Symptom | Guide |
|---|---|---|
| 3.2 → 4.0 | warns, silently wrong rows | 3a, rewritten |
| 4.0 → 4.1 | `NoMethodError`, lazily on first call; colliding names rejected | 7b, 7c |
| 4.1 → 4.2 | `ArgumentError` at class load | 5b, 5c |

**Guides.** Section 3a now carries the 3.2-vs-4.0 `#scope` source, the measured `to_sql`, a per-version behavior table, and an auditing order that starts with association call sites — the only places where the bug is observable. Plus troubleshooting rows for the symptoms as an operator actually meets them: a query returning the whole table, an unexplained 503, `ArgumentError: The scope body needs to be callable.`

**`CLAUDE.md`.** New `kind:` rule: *a deprecation that changes query results is `breaking` at the hop where the behavior changes*. The deciding question is whether a green suite and a clean boot can hide it, not whether Rails prints a warning. Without this, the rubric's rule 2 argues for downgrading the 4.0 entry to `deprecation`. Also corrected the section's example, which claimed this API was deprecated in 3.1 and raised in 4.0 — verified against the gems: 3.1 had no deprecation, and 4.0 warns.

## Verification

Behavior confirmed against the vendored activerecord 3.2.22.5 and 4.0.13 gems and the tagged v4.1.0 / v4.2.0 sources. The SQL was measured by running both boots of a real dual-boot app, not inferred.

Both regexes were run against a ~900-scope Packwerk codebase before and after its own sweep:

- scope bodies: 17 findings with the old regex, 23 with the new one, 0 false positives on the swept branch
- `default_scope`: 0 of 7 declarations flagged (all block form), after a `\b` guard was added — the first run matched a `default_scoper` helper call

Both false positives that run surfaced (a parenthesised `proc` body, `default_scoper`) are now fixtures. `bin/validate-patterns`, `bin/test-patterns` (both with `--self-test`) and `bin/lint-skill` pass.

## Note for reviewers

`SCOPE_WITHOUT_LAMBDA` stays `kind: breaking` at the 4.0 hop even though Rails 4.0 only warns. That is deliberate and the pattern entries carry inline comments saying so — the failure mode is wrong query results, which a passing test suite hides. `DEFAULT_SCOPE_WITHOUT_BLOCK` is also kept `breaking` at 4.0, where it is genuinely deprecated-but-correct, so that it gets fixed in the same pass as the scope bodies rather than sending someone back over the same models at 4.1.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -81,7 +81,11 @@ Priority is about **urgency during an upgrade**, not editorial weight.
 
 **Judge `kind` at the target hop, not the API's historical timeline.** Each `rails-XY-patterns.yml` file is a statement *about that hop* — what changes when the user upgrades INTO that version. A removal that was first deprecated in an earlier Rails minor is `breaking` in the file for the version where it actually raises, not `deprecation` because of its history. The same API can legitimately be `deprecation` in `rails-31-patterns.yml` and `breaking` in `rails-40-patterns.yml`. Apply the rule to all four kinds: `kind` reflects what the change *is at this hop*, not what it *was* earlier or *will become* later.
 
-Concrete example: `SCOPE_WITHOUT_LAMBDA` was deprecated in Rails 3.1 and raises in 4.0 — it is `breaking` in `rails-40-patterns.yml`.
+Concrete example: `SCOPE_WITHOUT_LAMBDA`. Rails 3.2 accepts a non-callable scope body and merges it into the current scope; 4.0 warns and hands the stored relation back verbatim, silently dropping the association or chained scope; 4.1 raises `NoMethodError` at call time; 4.2 raises `ArgumentError` at class load. It is `breaking` in `rails-40-patterns.yml`, `rails-41-patterns.yml`, and `rails-42-patterns.yml` — see the next rule for why the 4.0 entry is not a `deprecation`.
+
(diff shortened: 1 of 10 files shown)


JuanVqz:
This seems app-related, not general, so closing it

PR #130, opened 2026-08-26
Add remote-form CSRF token coverage for the 3.2 to 4.0 hop

JuanVqz:
Rails 4.0 flips `embed_authenticity_token_in_remote_forms` to `false`, so `remote: true` forms stop rendering the hidden `authenticity_token`. The token then only reaches the server as the `X-CSRF-Token` header UJS sets, so any submit path that bypasses UJS posts with no token. A bare `protect_from_forgery` nulls the session instead of raising, so the request silently loses its user.

Nothing warns and no test catches it: the Rails-side change is an absent config line, the failure is browser-only, and `allow_forgery_protection` is off in the test environment.


etagwerker (approved):
@JuanVqz Looks good, thanks!

PR #131, opened 2026-08-27
Add order / reorder no-arguments coverage for 4.0

JuanVqz:
## Summary

Adds coverage for `order` / `reorder` requiring arguments on Rails 4.0, hit during a real 3.2 → 4.0 release.

Rails 3.2 was `def order(*args); return self if args.blank?; ...`, so a bare `order` returned the relation unchanged. Rails 4.0 calls `check_if_method_has_arguments!` in both `order` and `reorder`, so `order` and `order()` raise `ArgumentError: The method .order() must contain arguments.` ([4-0-stable](https://github.com/rails/rails/blob/4-0-stable/activerecord/lib/active_record/relation/query_methods.rb)).

Only those two forms. The guard tests the splat array, so `order(nil)` passes `[nil]` and `order([])` passes `[[]]`, neither is `blank?`, and both keep working. `reorder(nil)` stays the way to clear a default order. The pattern leaves them alone.

It raises at runtime on the call site, not at boot, so a report or a worker that no spec exercises passes CI and raises in production.

## What this adds

- `ORDER_WITHOUT_ARGUMENTS` in `rails-40-patterns.yml`, with fixtures. The regex needs a relation method right after the bare call, ending there, so `line_item.order` and `line_item.order.summary` don't drown the real hits.
- Section 6 in `upgrade-3.2-to-4.0.md` (6–24 renumbered to 7–25), plus two rows in the error lookup table.

## Fixes to the existing hash pattern

`SQUEEL_ORDER_JOINED_TABLE` already covered the hash half, but:

- `search_paths` was `app/models/` only, so query objects, reports and workers were never scanned. Now `app/` + `lib/`.
- `exclude: ":asc|:desc"` is a substring match, so a `:description` column anywhere on the line silently dropped a real finding. Anchored.
- the regex missed `order({ table: :column })` and matched `work_order(status: :open)`. Braces added, left side anchored.
- explanation blamed Squeel. Squeel is one cause; the other is a call that was never sorting, since 3.2 serialized the hash into the `ORDER BY` string. That one matters more, because fixing it changes output that has been wrong since it was written.
- no fixtures. Added.

## Verification

- `bin/validate-patterns` and `bin/test-patterns` clean (177 assertions).
- Guide greps checked against the shipped regex on the same 6 lines, same result for each.
- `order(:id).last` ≡ 3.2 `.order.last`: 3.2 `last` with no order calls `reverse_order`, which falls back to the primary key.


etagwerker (approved):
@JuanVqz Looks good!

Issue #4, opened 2026-03-23
When I tell Claude Code to upgrade from X to Y, the methodology needs to be slightly tweaked


etagwerker:
Here is what the skill reports: 

```
  Per the upgrade methodology, I need to follow these steps:
  1. Run the test suite (mandatory)
  2. Verify load_defaults
  3. Run breaking changes detection
  4. Generate reports
```

I believe it should be slightly different: 

```
Per the upgrade methodology, I need to follow these steps: 
  1. Update the Gemfile to have the right values of X and Y. Then bundle install and next bundle install 
  2. Run `rails runner 'puts Rails.version'`
  3. Make sure that deprecation warnings are not silenced in test environment
  4. If deprecation warnings are silenced, changed configuration to output deprecation warnings in $stdout 
  5. Run the test suite (mandatory)
  6. Collect a list of deprecation warnings 
  7. Verify load_defaults
  8. Run breaking changes detection
  9. Address deprecation warnings raised in the test suite execution for Rails version X
  10. Run `BUNDLE_GEMFILE=Gemfile.next rails runner 'puts Rails.version'` 
  11.  Run the test suite for the models' test suite with BUNDLE_GEMFILE=Gemfile.next
  12. Generate reports
```

JuanVqz:
Thanks for raising this. After reviewing against the FastRuby.io methodology I agreed the step sequence is off. Proposing this Core Workflow, what you mention is between the same lines:

1. Run Test Suite
2. Verify Latest Patch Version
3. Resolve Deprecation Warnings
4. Review Ruby Compatibility
5. Set Up Dual-Boot
6. Run Breaking Changes Detection
7. Fix Broken Build
8. Smoke Test
9. Remove Dual-Boot
10. Align `load_defaults` (optional, post-upgrade)
11. Generate Reports

Feedback welcome.

cc @arielj

PR #136, opened 2026-09-19
Fold issue #4 into the upgrade flow: resolve deprecations first, boot both sides, models suite on the next side (v4.1.0)

JuanVqz:
## Why

Issue #4 asked for the methodology to make deprecation warnings visible, collect and fix the ones the current Rails version emits before touching the next version, boot both sides after the Gemfile change, and prove the app's models pass on the target before any report is written. The old flow ran tests, then detection, then reports, and left deprecation handling implicit.

This PR folds those items into the existing flow without changing its shape. It is the first methodology change on top of #134 and shows why that restructure was done: one new workflow file, one index row, two small additions to existing workflows, and a renumber. No template, checklist or principle list had to be edited in three places.

## What changed

| # | Workflow | Change |
|---|----------|--------|
| 00 | Verify latest patch | unchanged |
| 01 | Run test suite | unchanged; hands off to 02 |
| 02 | Resolve deprecation warnings | **new**: visibility (issue items 3 and 4), collect (6), triage and fix on the current version (9), inventory for the report |
| 03 | Validate upgrade path | unchanged |
| 04 | Set up next_rails | was 02; gate now boots both sides with `rails runner 'puts Rails.version'` (items 1, 2 and 10) |
| 05 | Detect breaking changes | was 04 |
| 06 | Check gem compatibility | was 05 |
| 07 | Boot smoke test and models suite | was 06; models suite under `Gemfile.next` **new** (item 11), failures go to the fix-before-bump bucket |
| 08 | Generate upgrade report | was 07 (item 12) |
| 09 | Generate app:update preview | was 08 |
| 10 | Implement and upgrade | was 09 |
| 11 | Sync CI config | was 10 |
| 12 | Align load_defaults | was 11 (item 7, kept after the bump per the PR #26 decision) |
| 13 | Mention cleanup | was 12 |

Not taken from the issue's list: dual-boot before the test suite (a green baseline comes first), and reports after implementation (the report is the plan the user reviews; the Reports Only request shape depends on it).

Workflow 02 fixes deprecations before dual-boot on purpose: every fix is an unconditional replacement the current version already accepts, so nothing needs a `NextRails.next?` branch and the later bump diff stays about the bump.

## Commit groups

- `move:` renumber, with every `Workflow NN` mention and path remapped in the same commit so lint stays green.
- `content:` the new workflow, the boot gate, the models suite step, the hand-off lines between renumbered neighbours, the index and request-patterns matrix. A tests-first swap and a Ruby-compatibility step were tried and reverted: they came from a later comment's alternative structure, not from the issue, and the goal here is a good base before new features.
- `lint:` version 4.1.0, CHANGELOG, CLAUDE.md example path.
- `fix:` what one review round and one QA run of the new flow against a real app found in the new text (see below).

## QA

Same harness as #134: a fresh agent ran the skill against rubymem.com (Rails 7.0.10 to 7.1) in its own worktree with the request "upgrade to 7.1, stop after both reports", logging every file read, step, stop and confusion. The new order executed as written: Workflow 02 inventoried the current-version warnings, Workflow 04's gate booted both sides, Workflow 07 ran the models suite under `Gemfile.next` (35 tests green), and the Reports Only shape stopped after 09. Every gate passed first try. 15 minutes, 216k tokens, 43 tool calls, in line with the runs on #134. Four confusions about the new text were fixed on this branch; the remaining twelve are the pre-existing pattern, template and version-guide defects already tracked from #134's QA.

## Verification

- `bin/lint-skill`, `bin/validate-patterns`, `bin/test-patterns` green after every commit.
- Every `workflows/NN-` path and `Workflow NN` mention across `SKILL.md`, workflows, references, templates and pattern YAML resolves to the new numbering.

Closes #4.


etagwerker (approved):
@JuanVqz Looks good!

